PR #7, opened 2026-04-15
Add base front-end files.

victor-paunescu:


ui/src/app/app.ts
```diff
@@ -0,0 +1,94 @@
...
+          this.configService.resetProjectConfig();
+          return;
+        }
+        let root = this.router.routerState.snapshot.root;
+        while (root.firstChild) {
+          root = root.firstChild;
+        }
+        const id = root.paramMap.get('id');
+        if (id !== null) {
+          if (this.loggedIn()) {
```
adamread, line 69: I recall forcing `loadProjectConfig()` needing to sometimes be in the `setTimeout()`, but I don't remember why. Can you add a comment here for future reference?

adamread (dismissed):
One nit about documenting something that looks like a hack.

Otherwise, if you're sure about the import order, LGTM

adamread (approved):
LGTM

PR #17, opened 2026-04-15
Py: Analyse file action

akushnarov:


actions/analyse_file.py
```diff
@@ -0,0 +1,72 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Analyses a file using Gemini."""
+
+from __future__ import annotations
+
+import json
+from typing import Any
```
adamread: nit: you shouldn't need to import from typing anymore. You can just use `dict` any `any` in your annotations directly (although you should avoid any...)

actions/analyse_file.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def execute(
+    gcs: GCS,
+    params: Params,
+    prompt: str,
+    file: NodeInput,
+    generation_prompt: NodeInput,
+    response_schema: Dict[str, Any],
+    gemini_model: str = 'gemini-2.5-flash',
```
adamread: nit: It might be prudent to remove the defaults for model versions moving forward to reduce the number of places they need to be changed when there are updates.

actions_lib/test/test_gemini.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+        self.assertEqual(output, "mock response")
+
+        # Verify interactions
+        mock_client_class.assert_called_once_with(
+            vertexai=True, project="test-proj", location="us-central1"
+        )
+
+    @mock.patch("google.genai.Client")
+    def test_prompt_json_schema(self, mock_client_class):
```
adamread, line 74: This seems to only test the mock. It might be good to add some checks to ensure the Gemini API call is correct.

actions_lib/gemini.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def prompt(
+    gcp_project: str,
+    text_prompt: str,
+    response_schema: Dict[str, Any] = None,
+    file_uris: List[str] = None,
+    need_to_remove_md_notation=True,
+    location="us-central1",
+    model="gemini-2.5-flash",
```
adamread, line 83: nit: see other remark about the default model.

actions_lib/gemini.py
```diff
@@ -0,0 +1,162 @@
...
+        top_p=top_p,
+        max_output_tokens=8192,
+        response_modalities=[types.Modality.TEXT.value]
+        if not response_schema
+        else None,
+        response_mime_type="application/json" if response_schema else None,
+        response_schema=response_schema,
+        safety_settings=safety_settings,
+        thinking_config=types.ThinkingConfig(thinking_budget=0)
+        if model != "gemini-2.5-pro"
```
adamread: This doesn't seem very future proof. Why can't the thinking_budget always just be 0?

adamread (commented):
Seems like this is all indented with 4 spaces. I don't think that's correct.

akushnarov:
> Seems like this is all indented with 4 spaces. I don't think that's correct.

According to our py style guide it's 4
https://google.github.io/styleguide/pyguide.html#s3.4-indentation

adamread:
The rest of the codebase seems to use two spaces. The CL is consistent, so it's not blocking.

actions/analyse_file.py
akushnarov: Implementing dict and specific typing. I think "any" cannot be used as a type, only "Any" (https://docs.python.org/3/library/functions.html#any)

actions/analyse_file.py
akushnarov: Makes sense

actions_lib/test/test_gemini.py
akushnarov: Implemented

actions_lib/gemini.py
akushnarov: Will leave for the future refactoring, since old parts of the code don't set those params

actions_lib/gemini.py
akushnarov: Added the checks for 2.5 models, the models starting from 3 use thinking_level, this is for the future implementation

actions_lib/test/test_gemini.py
```diff
@@ -109,5 +118,48 @@ def test_prompt_with_files(self, mock_client_class):
         self.assertEqual(output, "analyzed")
 
 
+    @mock.patch("google.genai.Client")
+    def test_prompt_thinking_config_pro(self, mock_client_class):
```
cahlheim: is it worth combining those two into a parameterized test?

cahlheim (approved):

actions_lib/test/test_gemini.py
akushnarov: Thanks, implemented

requirements-dev.txt
```diff
@@ -1,335 +1,68 @@
...
+# This file is autogenerated by pip-compile with Python 3.12
 # by the following command:
 #
-#    pip-compile --generate-hashes --no-emit-index-url --output-file=requirements-dev.txt requirements-dev.in
+#    pip-compile requirements-dev.in
 #
-annotated-types==0.7.0 \
-    --hash=sha256:1f02e8b43a8fbbc3f3e0d4f0f4bfc8131bcb4eebe8849b8e5c773f3a1c582a53 \
-    --hash=sha256:aff07c09a53a08bc8cfccb9c85b05f1aa9a2a6f23728d790723543408344ce89
+--index-url https://us-python.pkg.dev/artifact-foundry-prod/ah-3p-staging-python/simple/
```
victor-paunescu: Please compile the requirements with the same command as before
`pip-compile --generate-hashes --no-emit-index-url --output-file=requirements-dev.txt requirements-dev.in`

victor-paunescu (changes requested):

akushnarov (commented):
Done

victor-paunescu (approved):

PR #22, opened 2026-04-15
Add remix-engine service

victor-paunescu:


ui/src/app/services/remix-engine/remix-engine.interface.ts
```diff
@@ -0,0 +1,135 @@
...
+ * limitations under the License.
+ */
+
+import {AspectRatio, Resolution} from '../config/config';
+
+/**
+ * Represents a single scene item within a generated storyboard.
+ */
+export interface StoryboardItem {
+  product_id: string | number;
```
adamread: For here and below, I'm sure this works, but in what case do you use them as a number?

ui/src/app/services/remix-engine/remix-engine.interface.ts
```diff
@@ -0,0 +1,135 @@
...
+  scene_name: string;
+  video_prompt: string;
+}
+
+/**
+ * Represents a generic node item in the workflow output, extending a
+ * storyboard item.
+ */
+export interface NodeItem extends Partial<StoryboardItem> {
+  _error?: string;
```
adamread, line 34: Don't use leading underscores (go/tsjs-style#naming-style) unless this is coming from the API, in which case the python shouldn't be exposing it.

akushnarov (dismissed):

ui/src/app/services/remix-engine/remix-engine.ts
```diff
@@ -0,0 +1,959 @@
...
+      if (!workflowStatus.sink) {
+        throw new Error('Workflow completed without output');
+      }
+      const storyboardJsonFile =
+        workflowStatus.sink.output['0']['storyboard'][0]?.file;
+      if (!storyboardJsonFile) {
+        throw new Error('Storyboard JSON file not found');
+      }
+
+      let storyboardJson;
```
adamread: nit: You could make this const and move the catch down with the others where the type of the error is checked anyhow. Not sure why the extra catch is inserted into the existing try block.

ui/src/app/services/remix-engine/remix-engine.ts
```diff
@@ -0,0 +1,959 @@
...
+    if (this.configService.isGeneratedScene(scene)) {
+      const candidate = scene.candidates![scene.selectedCandidateIndex!];
+      const end = candidate.trim?.end ?? candidate.durationSeconds;
+      return end - skipTime;
+    }
+    if (this.configService.isProvidedVideoScene(scene)) {
+      const end = scene.trim?.end ?? scene.durationSeconds!;
+      return end - skipTime;
+    }
+    return 0;
```
adamread, line 839: Can it happen that the scene is neither generated nor provided? That seems more like an error condition that should be caught before calling this.

ui/src/app/services/remix-engine/remix-engine.ts
```diff
@@ -0,0 +1,959 @@
...
+        }
+      } else if (this.configService.isGeneratedScene(scene)) {
+        if (scene.candidates && scene.selectedCandidateIndex !== undefined) {
+          return scene.candidates[scene.selectedCandidateIndex].video?.path;
+        }
+      }
+      return false;
+    });
+    for (const scene of validScenes) {
+      let gcsVideoPath;
```
adamread, line 861: If you checked if whether the scene is generated or provided first, and assigned it to a variable, then this and the next two lines could be const and you'd remove the duplicate code for the cases.

adamread (dismissed):
I left some comments, but they're all nits and more about style than content.

ui/src/app/services/remix-engine/remix-engine.interface.ts
victor-paunescu: The API returns it like this, and we want it to be exposed.
We could look into changing the name in the API.

ui/src/app/services/remix-engine/remix-engine.interface.ts
victor-paunescu: I removed the | number.

ui/src/app/services/remix-engine/remix-engine.ts
victor-paunescu: Done. I added a SyntaxError check to catch the JSON Parse error.

ui/src/app/services/remix-engine/remix-engine.ts
victor-paunescu: No, at the moment we either have a generated or a provided scene.
We can look into adding additional checks in a future commit.

ui/src/app/services/remix-engine/remix-engine.ts
victor-paunescu: I removed duration from the if clauses.
Imo extracting the video into a variable wouldn't make it much clearer.

adamread: I dislike the repetition, but menh.

adamread (approved):
LGTM

PR #30, opened 2026-04-15
Add files referenced by demonstration workflows

uenke:
This is a collection of purpose-generated text, images, videos and sounds needed by the supplied demo workflows.


adamread (approved):
LGTM. We should ensure other people are using these as test fixtures.

PR #36, opened 2026-04-16
Adding the generate component for the UI.

adamread:
This adds the generate component, that creates the project UUID and redirects.


victor-paunescu (approved):
As discussed, in the future we can move this either to a service, if needed by multiple components, or to the homepage component.

PR #37, opened 2026-04-16
Outpainter library and tests

akushnarov:


actions_lib/outpainter.py
```diff
@@ -0,0 +1,98 @@
...
+
+import io
+from typing import Tuple
+
+from google import genai
+from google.genai import types as gtypes
+import PIL.Image
+
+
+IMAGE_MODEL = "gemini-2.5-flash-image"
```
adamread: Shouldn't this come from a config or be passed into the function?

adamread (dismissed):
nit about the model being hard coded and, of course, the indentation =P

actions_lib/outpainter.py
akushnarov: 1. The current library supports only `gemini-2.5-flash-image` params. If someone specifies 3.1 it won't work
2. We don't have a separate config variable for Nano Banana api version

caiotomazelli: use _CAPS_WITH_UNDER format for internal global variables

caiotomazelli (approved):

akushnarov (commented):
Done

adamread (approved):
LGTM, although you could have fixed the indentation.

PR #40, opened 2026-04-16
Py: generate storyboard action

akushnarov:


actions/generate_storyboard.py
```diff
@@ -0,0 +1,418 @@
...
+        ],
+      },
+    }
+  },
+  "required": ["storyboard"],
+}
+
+
+def generate_system_prompt(min_scenes: int = 1, max_scenes: int = 3) -> str:
+  """Return a system prompt for the generate_storyboard action.
```
AndreyMarkinPPC: Nit: The docstring may be descriptive-style ("""Fetches rows from a Bigtable.""") or imperative-style ("""Fetch rows from a Bigtable."""), but the style should be consistent within a file. https://google.github.io/styleguide/pyguide.html#383-functions-and-methods.
So better to have it as `Returns`.

actions/generate_storyboard.py
```diff
@@ -0,0 +1,418 @@
...
+
+  client = genai.Client(
+      vertexai=True,
+      project=workflow_params[Key.GCP_PROJECT.value],
+      location=gemini_model_location,
+  )
+
+  config = genai.types.GenerateContentConfig(
+      system_instruction=generate_system_prompt(
+          len(products.keys()), len(images)
```
AndreyMarkinPPC: Nit: `len(dict.keys())` is the same as `len(dict)`

AndreyMarkinPPC (approved):

actions/generate_storyboard.py
akushnarov: Thanks! It's also much faster:
Time for len(dict):      0.000281 seconds
Time for len(dict.keys()): 0.000463 seconds
Ratio (keys / len):      1.65x

actions/generate_storyboard.py
akushnarov: Good catch

victor-paunescu (approved):

PR #43, opened 2026-04-16
Py: Image describer with test

akushnarov:


AndreyMarkinPPC (approved):
By the way, a shameful plug here - https://google.github.io/filonov/tagging/overview/. I have a library this is supposed to do such a description generic.  You can provide a custom prompt and a schema and then it handles the rest.

victor-paunescu (approved):

actions_lib/image_describer.py
```diff
@@ -0,0 +1,322 @@
...
+      Action within the Shot: Use the shot.action field to describe the overall event happening within the frame (e.g., "A car is driving down the street," "Rain is falling"). If the scene is static and no discernible event is occurring, leave this field as an empty string "".
+      Grounded Technical & Interpretive Analysis: For fields requiring technical analysis (shot, cinematography) or subjective interpretation (mood, tone), your response must be strictly grounded in visual evidence. Interpretive terms (e.g., "Nostalgic," "Soft Light") are acceptable only if they are a direct and logical interpretation of the image's lighting, composition, and color palette.
+      Strict Adherence to Schema: You must populate every single field in the provided JSON schema. If information for a field is not present or not applicable (per the rules above), you must use an empty string "". Do not omit any keys from the final JSON output.
+      {guidance_prompt}
+      </INSTRUCTIONS>
+    """
+  response_schema = {
```
jmistral-google, line 64: This is a static string, so I would extract into a module level internal constant:

```python
 _RESPONSE_SCHEMA_ = {
  ..
}
```

actions_lib/image_describer.py
```diff
@@ -0,0 +1,322 @@
...
+  Returns:
+    a JSON object with a description of the image.
+  """
+  if guidance:
+    guidance_prompt = (
+        f"\n\nThis identifies the focus object of the image: {guidance}"
+    )
+  else:
+    guidance_prompt = ""
+  prompt = f"""
```
jmistral-google, line 43: I would extract this too as an internal constant, then use the python `string.Template` to inject values.  This makes the function much more readable.

```python
_IMAGE_ANALYSIS_PROMPT = """
...
"""

prompt_template = string.Template(_IMAGE_ANALYSIS_TEMPLATE)
prompt = prompt_template.substitue(
  guidance = ...,
)
...
```

PR #44, opened 2026-04-16
Added write product script action and tests

akushnarov:


actions/write_products_script.py
```diff
@@ -0,0 +1,277 @@
...
+    image_description: The image descriptions.
+    story_variant_quantity: The number of variants to generate.
+    gemini_model: The Gemini model to use.
+    gemini_model_location: The model's location.
+
+  Returns:
+    A NodeOutput object containing the ad script.
+
+  Raises:
+    RuntimeError: for repeated invalid outputs.
```
AndreyMarkinPPC: Nit: should be capitalized `For`

akushnarov: Done

actions/write_products_script.py
```diff
@@ -0,0 +1,277 @@
...
+                  ],
+              },
+          },
+      },
+      "required": ["style", "scenes"],
+  }
+  logger.debug("Prompt: %s", text_prompt)
+  for story_variant_id in range(story_variant_quantity):
+    result = None
+    for _ in range(4):  # 1 initial call + 3 retries
```
AndreyMarkinPPC, line 203: Super nit suggestion - you can combined `tenacity` with `pydantic` validation to avoid writing validation / retries by hand (maybe not in this particular case).
https://tenacity.readthedocs.io/en/latest/

You apply `retry` decorator and tell how many type and on what exceptions you want to retry.

```
@tenacity.retry(
    stop=tenacity.stop_after_attempt(3),
    retry=tenacity.retry_if_exception_type(pydantic.ValidationError)
)
def method_name(self, ...)
```

thus you decouple retry logic from the business logic.

AndreyMarkinPPC (approved):

AndreyMarkinPPC:
I noticed that that you have tests but don't run them when PR is created. 
Is this done on purpose?
Anyway it would be great to get this feedback as a part of PR review.

akushnarov:
> I noticed that that you have tests but don't run them when PR is created. Is this done on purpose? Anyway it would be great to get this feedback as a part of PR review.

I guess we overlooked that. Implementing https://github.com/google-marketing-solutions/scene-machine/pull/53

victor-paunescu (approved):

PR #50, opened 2026-04-17
Finalise user-agent schema; Fix linting issues

uenke:


actions_lib/gemini.py
```diff
@@ -19,144 +19,154 @@
...
-    response_schema: Dict[str, Any] = None,
-    file_uris: List[str] = None,
+    response_schema: Dict[str, Any] | None = None,
+    file_uris: List[str] | None = None,
     need_to_remove_md_notation=True,
     location="us-central1",
     model="gemini-2.5-flash",
     temperature: float = 0.2,
     top_p: float = 0.2,
+    call_type: TrackingType | None = None,
```
akushnarov: nit: I would call it tracking_type to be consistent with the type

common.py
```diff
@@ -116,17 +126,21 @@ class ContentType(enum.Enum):
...
 
+  Args:
+    context: The string to denote what type of request is being made.
+
   Returns:
     A dictionary of headers.
   """
-  user_agent = 'gtech-ads-ce/scene-machine/'
+
+  user_agent = _USER_AGENT_PREFIX + '-' + str(context) + '-v'
```
akushnarov: *Issue*: Since context is typed as an Enum (TrackingType), calling str(context) will evaluate to the string "TrackingType.VIDEO" rather than the underlying value "video". This will result in an incorrectly formatted user agent like cloud-solutions/mas-scenemachine-TrackingType.VIDEO-v....

*Fix*: Use .value to extract the string:
user_agent = f"{_USER_AGENT_PREFIX}-{context.value}-v"

actions_lib/veo.py
```diff
@@ -43,12 +44,37 @@ def generate(
...
+  Raises:
+    RuntimeError: If no videos were generated.
+  """
   client = genai.Client(
       vertexai=True,
       project=gcp_project,
       location=gcp_location,
-      http_options=types.HttpOptions(headers=get_api_client_headers()),
+      http_options=types.HttpOptions(
+          headers=get_api_client_headers(TrackingType.VIDEO.value)
```
akushnarov: Issue: get_api_client_headers expects TrackingType not a string

fix: get_api_client_headers(TrackingType.VIDEO)

actions_lib/gemini.py
```diff
@@ -19,144 +19,154 @@
 import mimetypes
 from typing import Any, Dict, List
```
akushnarov: nit: you shouldn't need to import from typing anymore. You can just use dict & list in your annotations directly (although you should avoid any) 

From this comment: https://github.com/google-marketing-solutions/scene-machine/pull/17#discussion_r3091616157

actions_lib/gemini.py
```diff
@@ -19,144 +19,154 @@
...
+  """Removes the json wrapper from a Gemini response."""
+  return s.replace("```json", "").replace("```", "")
 
 
 def prompt(
     gcp_project: str,
     text_prompt: str,
-    response_schema: Dict[str, Any] = None,
-    file_uris: List[str] = None,
+    response_schema: Dict[str, Any] | None = None,
```
akushnarov: You can do something like: 
`
response_schema: dict[
        str, str | dict | list | int | float | bool
    ] = None,
`

common.py
```diff
@@ -116,17 +126,21 @@ class ContentType(enum.Enum):
 
 
 @functools.lru_cache(maxsize=1)
```
akushnarov: Issue: Previously, this function took no arguments, so maxsize=1 was appropriate. Now that it accepts a context argument, a maxsize=1 will result in caching only one last call/param.

Fix: Increase the cache size to accommodate all tracking types, or simply set it to None:
@functools.lru_cache(maxsize=None)

akushnarov (changes requested):

actions_lib/gemini.py
uenke: I think then there are complaints about the inner dict or list not having arguments.

actions_lib/gemini.py
uenke: Consistent with which type? The style guide says "Use CapWords for class names", and this is a class in common.py. Besides, the other classes there are named similarly, so changing this would be a major undertaking.

actions_lib/gemini.py
uenke: Acknowledged – I even removed the typing types in some places, except Any (see below). But this is the file you committed, I didn't touch these aspects.

actions_lib/veo.py
uenke: Thanks, done!

common.py
uenke: Done.

common.py
uenke: Done.

actions_lib/gemini.py
uenke: Oh, sorry, you meant the variable name. Right.

akushnarov:
The code is in the different branch: https://github.com/google-marketing-solutions/scene-machine/pull/63

PR #56, opened 2026-04-21
Upgrade py modules

akushnarov:


victor-paunescu:
I think pytest is missing in the new requirements-dev.txt

victor-paunescu (approved):

PR #60, opened 2026-04-21
Only allow the project owner to delete a project

olegrakkolainen:
The current setup for deleting projects allows everyone to delete everyone's project. This change makes it so that only the project owner can delete a project.

### DIFF
--- a/ui/src/app/setup/setup.html
+++ b/ui/src/app/setup/setup.html
@@ -385,6 +385,10 @@
           title="Delete Project"
           [disabled]="userEmail() !== createdByEmail()"
           (click)="deleteProject()"
+          matTooltipPosition="above"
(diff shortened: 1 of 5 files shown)


ui/src/app/setup/setup.html
```diff
@@ -383,6 +383,7 @@ <h2><span class="optional-info">Optional</span><br />Composition</h2>
           mat-flat-button
           class="delete-project-btn"
           title="Delete Project"
+          [disabled]="userEmail() !== createdByEmail()"
```
victor-paunescu, line 386: nit: could also add a tooltip explaining why you can't delete, in a future commit

victor-paunescu (dismissed):

victor-paunescu (approved):

PR #63, opened 2026-04-23
User agent for Vertex API calls

akushnarov:


common.py
```diff
@@ -115,16 +115,32 @@ class ContentType(enum.Enum):
...
+class TrackingType(enum.Enum):
+  """Enumerates tracking types for user-agent logging."""
+
+  VIDEO = 'video'
+  STORYBOARD = 'storyboard'
+  IMAGE = 'image'
+  PROMPT = 'prompt'
+
+
+_USER_AGENT_PREFIX = 'gtech-ads-ce/scene-machine'
```
AndreyMarkinPPC, line 127: Do you want to hardcode it here? Maybe provide it as ENV variable?

AndreyMarkinPPC (approved):

victor-paunescu (approved):

PR #65, opened 2026-04-24
Add a login button

victor-paunescu:
Fixes an issue where users with blocked popups would not be able to login.


akushnarov (approved):
nit:

PR #68, opened 2026-04-24
Configurable outpainter model

akushnarov:


actions/test/test_outpaint_image.py
```diff
@@ -57,7 +57,9 @@ def test_execute_correct_ratio(self, mock_outpainter_func):
         self.mock_gcs,
         self.mock_workflow_params,
         self.mock_image,
-        target_ratio
+        target_ratio,
+        'gemini-3.1-flash-image-preview',
```
adamread, line 61: nit: if the value doesn't matter, it's often better to make that clear in the test. e.g. I'd use `some-gemini-model` here to make it clear we don't depend on this specific model string.

There are a bunch of these, so I'll only mention it here.

adamread (dismissed):
One nit in the tests that can be addressed in future PRs.

actions_lib/outpainter.py
```diff
@@ -33,12 +33,12 @@
...
-
-_IMAGE_MODEL = "gemini-2.5-flash-image"
-
-
 def outpaint_image(
-    image_bytes: bytes, gcp_project: str, gcp_location: str, target_ratio: str
+    image_bytes: bytes,
+    gcp_project: str,
+    gcp_location: str,
+    outpainter_model: str,
```
evil-shrike, line 40: please update the docstring for the new argument

victor-paunescu (dismissed):

adamread (approved):
LGTM

PR #73, opened 2026-04-29
Image conversion

olegrakkolainen:
Modify old "ThumbnailService" to also cover cases for image conversion, and rename to "ClientMediaService". Additionally, block users from uploading image types other than JPEG, PNG, AVIF or WEBP.

### DIFF
--- a/ui/src/app/services/client-media/client-media.ts
+++ b/ui/src/app/services/client-media/client-media.ts
@@ -17,13 +17,15 @@
 import {Injectable} from '@angular/core';
 
 /**
- * Service for generating thumbnails from video URLs on the frontend.
+ * Service for transforming media using canvas, e.g. for generating thumbnails.
  */
 @Injectable({
   providedIn: 'root',
 })
-export class GenerateThumbnailService {
+export class ClientMediaService {
   constructor() {}
+
+  private readonly defaultMimeType = 'image/jpeg';
 
   private canvasFromMedia(
     src: HTMLImageElement | HTMLVideoElement,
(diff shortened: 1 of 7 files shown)


ui/src/app/services/client-media/client-media.ts
```diff
@@ -76,26 +78,17 @@ export class GenerateThumbnailService {
     return canvas;
   }
 
-  /**
```
victor-paunescu, line 79: Any reason to remove the docs instead of updating?

ui/src/app/services/client-media/client-media.ts
```diff
@@ -167,13 +166,22 @@ export class GenerateThumbnailService {
...
       imageElement.onload = () => {
         try {
-          const canvas = this.canvasFromMedia(imageElement, options);
-          canvas.toBlob(blob => {
-            if (!blob) {
-              throw new Error('Failed to generate thumbnail');
-            }
-            resolve(blob);
+          const canvas = this.canvasFromMedia(imageElement, {
+            maxWidth: options.maxWidth,
```
victor-paunescu, line 170: Do we need to check for cases where maxWidth and maxHeight are missing in options?

ui/src/app/services/client-media/client-media.ts
olegrakkolainen: I noticed that we don't generally do docs, so this _seems_ to be more inline. But happy to put them back and update instead.

ui/src/app/services/client-media/client-media.ts
olegrakkolainen: Not really as then they just won't be applied (maxWidth and maxHeight, when exist, are used to restrict output images to those sizes). But let me know if you see potential issues with this and / or how to better handle this!

victor-paunescu: If they don't get applied when undefined that's fine

victor-paunescu (approved):

PR #78, opened 2026-04-30
docs: update navigation links and add Google Cloud API list

adamread:
Fixed an issue where the header links were broken.

Also added a list of all the Cloud APIs used by Scene Machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,10 @@ _Disclaimer: This is not an officially supported Google product._
 
 In the generation steps, the user has full control, but can also rely on the tool's recommendations.
 
-[Technical Requirements](#requirements) •
+[Technical Requirements](#technical-requirements) •
 [Deployment](#deployment) •
 [Using Scene Machine](#using-scene-machine) •
-[Alternatives](#alternatives) •
+[Alternatives](#alternatives-to-scene-machine) •
 [Developers' Guide](DEVELOPING.md)
 
 ## Technical Requirements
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,6 +42,28 @@ To deploy this application, you need a **project on Google Cloud Platform withou
 
 Scene Machine sends workflow definitions to Remix Engine, which orchestrates its functional modules (e.g. turning images into videos) and reports back on results.
 
+### Google Cloud APIs Used
+
+The following APIs are used by Scene Machine:
+
+- Vertex AI API ( aiplatform.googleapis.com ): Used for accessing Gemini and Veo models for text, image, and video generation.
```
christophervoelpel: 1) Should we name it "Gemini Enterprise Agent Platform (formerly Vertex AI)"
2) We still are not providing links to look up the latest Gemini, Veo - Let's add https://ai.google.dev/gemini-api/docs/models as a sub-bullet to "look up the latest Gemini and Veo model versions"

README.md
christophervoelpel: Is this [!TIP] actually well formatted? It looks a bit "off" when rendered. Might need another empty line

README.md
```diff
@@ -42,6 +42,28 @@ To deploy this application, you need a **project on Google Cloud Platform withou
...
+- API Gateway API ( apigateway.googleapis.com ): Used to create and manage the API Gateway that routes traffic to the backend.
+- Artifact Registry API ( artifactregistry.googleapis.com ): Used to store the Docker container images for the backend service.
```
christophervoelpel, line 51: P2: should we format the urls as links  ?

christophervoelpel (commented):
Question about calling Vertex by its new name & actually providing a link to Gemin/Veo api documents

One nit about formating urls as links

README.md
adamread: They aren't links. They are the official names of the APIs when you use them. Following googleapis.com links gets you a 404.

README.md
adamread: 1) They say naming things is the second hardest problem in CS... fixed.
2) We are. The link is on line 97 with the instructions on updating the config file.

README.md
adamread: Seems like they can't be indented. Fixed.

adamread:
> Question about calling Vertex by its new name & actually providing a link to Gemin/Veo api documents
> 
> One nit about formating urls as links

Changed the API name, commented on the link being further down the page, and the URLs are URIs, not URLs.

christophervoelpel (approved):
discussed links offline

PR #82, opened 2026-05-05
FFmpeg filter fix

adamread:
Small fix in remix-engine.ts to ensure candidate durations are correctly calculated when a candidate includes a trim.

Also added thorough unit tests for remix-engine.ts.

### DIFF
--- a/ui/src/app/services/remix-engine/remix-engine.spec.ts
+++ b/ui/src/app/services/remix-engine/remix-engine.spec.ts
@@ -21,8 +21,7 @@ import {EnvironmentInjector} from '@angular/core';
 import * as storage from '@angular/fire/storage';
 import {Storage} from '@angular/fire/storage';
 import {MatSnackBar} from '@angular/material/snack-bar';
-import {of, take, filter, repeat, defer} from 'rxjs';
-import {TestScheduler} from 'rxjs/testing';
+import {of} from 'rxjs';
 import {beforeEach, describe, expect, it, vi} from 'vitest';
 import {ClientMediaService} from '../client-media/client-media';
 import {ConfigService} from '../config/config';
@@ -217,43 +216,73 @@ describe('RemixEngineService', () => {
       expect(httpClientMock.get).toHaveBeenCalled();
     });
 
-    it('should poll until sink output is defined', () => {
-      const testScheduler = new TestScheduler((actual: any, expected: any) => {
-        expect(actual).toEqual(expected);
+    it('should poll until sink output is defined', async () => {
+      vi.useFakeTimers();
+      const mockResponse1 = {sink: {output: undefined}};
+      const mockResponse2 = {sink: {output: {'0': {video: []}}}};
+
+      let callCount = 0;
+      httpClientMock.get.mockImplementation(() => {
+        callCount++;
+        if (callCount === 1) {
+          return of(mockResponse1);
+        } else {
+          return of(mockResponse2);
+        }
       });
 
-      testScheduler.run(({expectObservable, cold}: any) => {
-        const mockResponse1 = {sink: {output: undefined}};
-        const mockResponse2 = {sink: {output: {'0': {video: []}}}};
+      const pollPromise = service.pollWorkflow(
+        'mock-execution-id',
+        'mock-project-id',
+      );
 
-        httpClientMock.get
-          .mockReturnValueOnce(cold('(a|)', {a: mockResponse1}))
-          .mockReturnValueOnce(cold('b', {b: mockResponse2}));
+      // Advance timers by 1ms to let timer(0, ...) emit
+      await vi.advanceTimersByTimeAsync(1);
 
-        // Recreate the pipe in the test to test it with marbles
-        const poll$ = defer(() =>
-          (service as any).getWorkflowStatus('mock-execution-id'),
-        ).pipe(
-          repeat({delay: 10}),
-          take(5),
-          filter((response: any) => response.sink?.output !== undefined),
-          take(1),
-        );
+      // First poll should happen immediately
+      expect(callCount).toBe(1);
 
-        expectObservable(poll$).toBe('----------(b|)', {b: mockResponse2});
-      });
-    });
+      // Advance timers by 15ms to trigger the next poll (interval is 10ms)
+      await vi.advanceTimersByTimeAsync(15);
+
+      const result = await pollPromise;
+
+      expect(result).toEqual(mockResponse2);
+      expect(callCount).toBe(2);
+
+      vi.useRealTimers();
+    }, 10000);
 
     it('should throw ProjectChangedError if project changes', async () => {
+      vi.useFakeTimers();
       const mockResponse = {sink: {output: undefined}};
-      httpClientMock.get.mockReturnValue(of(mockResponse));
+      httpClientMock.get.mockImplementation(() => {
+        console.log('Test 2: Mock HTTP GET called');
+        return of(mockResponse);
+      });
       configServiceMock.projectConfig.value.mockReturnValue({
         id: 'different-project-id',
       });
 
-      await expect(
-        service.pollWorkflow('mock-execution-id', 'mock-project-id'),
-      ).rejects.toThrow('Project changed');
+      const pollPromise = service.pollWorkflow(
+        'mock-execution-id',
+        'mock-project-id',
+      );
+      pollPromise.catch(() => {}); // Prevent unhandled rejection warning
+
+      // Advance timers by 1ms to let timer(0, ...) emit
+      await vi.advanceTimersByTimeAsync(1);
+
+      let thrownError: any;
+      try {
+        await pollPromise;
+      } catch (e) {
+        thrownError = e;
+      }
+
+      expect(thrownError).toBeTruthy();
+      expect(thrownError.message).toContain('Project changed');
+      vi.useRealTimers();
     });
   });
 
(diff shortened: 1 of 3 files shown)


ui/src/app/services/remix-engine/remix-engine.spec.ts
```diff
@@ -14,36 +14,1390 @@
...
+        const mockResponse1 = {sink: {output: undefined}};
+        const mockResponse2 = {sink: {output: {'0': {video: []}}}};
+
+        httpClientMock.get
+          .mockReturnValueOnce(cold('(a|)', {a: mockResponse1}))
+          .mockReturnValueOnce(cold('b', {b: mockResponse2}));
+
+        // Recreate the pipe in the test to test it with marbles
+        const poll$ = defer(() =>
+          (service as any).getWorkflowStatus('mock-execution-id'),
```
victor-paunescu: Why cast to any?

ui/src/app/services/remix-engine/remix-engine.spec.ts
```diff
@@ -14,36 +14,1390 @@
...
+        ).pipe(
+          repeat({delay: 10}),
+          take(5),
```
victor-paunescu: Why take 5 here?

ui/src/app/services/remix-engine/remix-engine.spec.ts
```diff
@@ -14,36 +14,1390 @@
...
+              video: [{file: 'mock-path/video.mp4'}],
+            },
+          },
+        },
+      };
+      vi.spyOn(service, 'pollWorkflow').mockResolvedValue(
+        mockWorkflowStatus as any,
+      );
+
+      (storage.getDownloadURL as any).mockResolvedValue('http://mock-url');
```
victor-paunescu: Why cast to any?

ui/src/app/services/remix-engine/remix-engine.spec.ts
```diff
@@ -14,36 +14,1390 @@
...
+      vi.spyOn(service, 'uploadThumbnail').mockResolvedValue({
+        path: 'mock-thumb-path',
+        url: 'mock-thumb-url',
+      });
+
+      await service.generateCandidates(mockScene as any, {
+        durationSeconds: 5,
+        model: 'model-1',
+        generateAudio: false,
+        resolution: 'LANDSCAPE' as any,
```
victor-paunescu: resolution should be of type Resolution ('720p' | '1080p' | '4k')

ui/src/app/services/remix-engine/remix-engine.spec.ts
```diff
@@ -14,36 +14,1390 @@
...
+        return newIds;
+      });
+
+      vi.spyOn(service, 'startVideoGenerationWorkflow');
+
+      await service.generateCandidates(mockScene as any, {
+        durationSeconds: 5,
+        model: 'model-1',
+        generateAudio: false,
+        resolution: 'LANDSCAPE' as any,
```
victor-paunescu: same here

ui/src/app/services/remix-engine/remix-engine.spec.ts
```diff
@@ -14,36 +14,1390 @@
...
+      );
+
+      expect(matSnackBarMock.open).toHaveBeenCalledWith(
+        'Failed to generate storyboard. Failed to start workflow',
+        'Dismiss',
+        {panelClass: ['error-snackbar']},
+      );
+    });
+
+    it('should handle workflow execution error and show error snackbar', async () => {
```
victor-paunescu, line 598: nit: storyboard workflow execution error.

I was confused how this was different from the test in L378

ui/src/app/services/remix-engine/remix-engine.spec.ts
```diff
@@ -14,36 +14,1390 @@
...
+      );
+
+      expect(matSnackBarMock.open).toHaveBeenCalledWith(
+        'Failed to generate storyboard. Mock storyboard error',
+        'Dismiss',
+        {panelClass: ['error-snackbar']},
+      );
+    });
+
+    it('should handle workflow completed without output and show error snackbar', async () => {
```
victor-paunescu, line 631: same here

ui/src/app/services/remix-engine/remix-engine.spec.ts
```diff
@@ -14,36 +14,1390 @@
...
+      );
+      expect(configServiceMock.addRenderRun).toHaveBeenCalledWith(
+        expect.objectContaining({
+          errorMessage: 'Workflow completed without output',
+        }),
+      );
+    });
+  });
+
+  describe('startStoryboardWorkflow', () => {
```
victor-paunescu, line 1214: Aren't these repeats of the tests we have above for startStoryboardWorkflow? 
If not, shouldn't we move the ones above to this section?

victor-paunescu (dismissed):

ui/src/app/services/remix-engine/remix-engine.spec.ts
adamread: It acts as a stop if there's a bug and the filter is wrong, the repeat above will cause the test to poll forever (which it did in a first attempt). The 5 will limit the polling to 5 attempts.

ui/src/app/services/remix-engine/remix-engine.spec.ts
adamread: `getWorkflowStatus` is private. TS won't let us call it otherwise. 

Looking at it again though, this is a bad test, since we aren't actually testing the implemented polling logic. This is the test that Gemini couldn't make work. Can I leave it here an refactor it manually in a new PR?

ui/src/app/services/remix-engine/remix-engine.spec.ts
adamread: I told Gemini to use dummy values where it could to ensure that it was clear that the value didn't matter. Of course, when using literal types, that breaks things without the cast. I'll update to just use 720p.

ui/src/app/services/remix-engine/remix-engine.spec.ts
adamread: same as above.

ui/src/app/services/remix-engine/remix-engine.spec.ts
adamread: I'm confused about the comment... The test above tests when the storyboard workflow returns undefined, meaning it never started. This one handles an error during execution, which returns the object with the `_error` attribute.

ui/src/app/services/remix-engine/remix-engine.spec.ts
adamread: also same here. These are all separate error paths in the front end, which show different error messages.

ui/src/app/services/remix-engine/remix-engine.spec.ts
adamread: No. The ones above are the result of calling `generateStoryboard`. These all call `startStoryboardWorkflow`, so errors are propagated instead of being shown to the user.

Honestly, I'm not happy with the way this works either. I think `startStoryboardWorkflow` should likely be private, since it's only called from `generateStoryboard`.

ui/src/app/services/remix-engine/remix-engine.spec.ts
adamread: old pattern. I've updated this and other casts to use `vi.mocked`

adamread (commented):
Thanks. I've made some updates.

ui/src/app/services/remix-engine/remix-engine.spec.ts
adamread: actually, there was a bug in the polling logic, where if the first emitted object is empty the takefirst fails and repeat doesn't trigger. I started with the correct test now and updated the logic to ensure it always retrys.

ui/src/app/services/remix-engine/remix-engine.spec.ts
victor-paunescu: you can ignore my comment. It was about the name of the test... but I hadn't seen it's under different describes

ui/src/app/services/remix-engine/remix-engine.spec.ts
victor-paunescu: it's fine for now. We can change it in a future PR

ui/src/app/services/remix-engine/remix-engine.ts
```diff
@@ -100,15 +109,16 @@ export class RemixEngineService {
     projectId: string,
   ): Promise<WorkflowStatusResponse> {
     return await firstValueFrom(
-      this.getWorkflowStatus(workflowId).pipe(
+      timer(0, WORKFLOW_STATUS_POLL_INTERVAL_MS).pipe(
```
victor-paunescu, line 112: why was this change needed?

victor-paunescu (approved):

ui/src/app/services/remix-engine/remix-engine.ts
adamread: It the request didn't complete (e.g. lost network) or errors out the repeat didn't trigger. Using the timer provides a driver for polling that we know will work independent of the state of the network request. Using switchMap avoids piling up if the backend responds more slowly than the timeout by abandoning the old request before sending the next.

adamread (commented):
Thanks

PR #88, opened 2026-05-11
deploy: harden prereqs, IAM retry, phase markers, manual-step UX

christophervoelpel:
Reworks `deploy.sh` and `deploy-ui.sh` to be friendlier on fresh GCP projects and harder to misuse. Captured during a fresh deployment to a standalone GCP project; full per-issue write-up below.

## Summary

| # | Improvement | Files |
|---|---|---|
| 1 | Pre-flight tool & auth checks | `deploy.sh` |
| 2 | Project-mismatch sanity check (gcloud active vs config.txt) | `deploy.sh` |
| 3 | `add_iam_binding` helper: suppresses verbose YAML, retries on etag conflicts with backoff, labels role | both |
| 4 | Block on manual prereqs (polling loops + non-TTY guards) | `deploy-ui.sh` |
| 5 | Guarantee default Compute Engine SA exists (enable compute.googleapis.com + wait) | `deploy.sh` |
| 6 | Warm up Vertex AI service agent (eliminates first-Veo "Service agents being provisioned" failure) | `deploy.sh` |
| 7 | `[▶]` phase markers — Ctrl+F bridge between terminal output and script source | both |
| 8 | End-of-script success banners with deep links + ready-to-paste gcloud snippet | both |
| 9 | README updates (Firebase Storage two-step, parallel-execution note, Adding Users section) | `README.md` |

## Why these specific changes

1. **Pre-flight checks** — running into a missing `envsubst` or unauthenticated gcloud session 30 seconds into a multi-minute call wastes time. Fail fast.
2. **Project-mismatch check** — caught in testing: easy to update `config.txt` but forget `gcloud config set project`, deploying to the wrong project. Now blocks with y/N.
3. **IAM helper** — original loop printed a full IAM policy YAML per binding (very noisy) and hit etag conflicts on fresh projects when background GCP work modified the policy in parallel. New helper suppresses output, retries with backoff per gcloud's own recommendation, and identifies the role in retry messages.
4. **Polling vs. exit-1** — original script `exit 1`'d on every unmet manual prereq, forcing re-runs and re-validation of everything from scratch. Each iteration: configure one thing in console → re-run → fail on next thing → repeat. Three prereqs = three re-runs minimum. Polling lets the user complete all manual steps once in one session.
5. **Compute SA guarantee** — `compute.googleapis.com` wasn't explicitly enabled. On some fresh projects the default Compute Engine SA isn't there yet at the time of role bindings, causing confusing "service account not found" errors. Adding it to the enable list + a wait loop closes the race.
6. **Vertex AI warm-up** — well-documented "Service agents are being provisioned" failure on the user's first Veo generation, caused by ~5–15 min propagation lag of the auto-granted Storage Object access. Triggering identity creation in the deploy + explicitly binding `roles/storage.objectUser` to the agent eliminates the failure window.
7. **Phase markers** — the deploy is long and chatty (~17 minutes, many gcloud subcommands). When something fails, mapping terminal output back to the right place in the script was painful. `[▶]` is a visually distinctive prefix that's also independently grep-able (square-bracket-glyph won't collide with anything else in the script source or terminal output).
8. **Final UX** — original final block printed the deploy URL inline with the IAP and "grant user access" warnings; URL was easy to miss; "Grant User Access" had no link, no console URL, no `gcloud` snippet.

## Notes

- **OAuth consent screen verification** uses a manual gate ("Press Enter when configured") rather than upstream verification, because `gcloud iap oauth-brands list` (the original verification) was permanently shut down on 2026-03-19 and required project-in-org anyway. There's no public API to verify OAuth-consent state on standalone projects. The gate has a non-TTY guard so CI contexts fail cleanly rather than silently auto-proceeding.
- All `[▶]` markers and the convention comment are documented inline in both scripts.
- Rebased on top of #90 (CUSTOM_DOMAIN); a one-line echo confirms when CUSTOM_DOMAIN is honored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,34 @@ The following APIs are used by Scene Machine:
 
 _Please note that most of the APIs are enabled automatically when you run the deployment script. Cloud Storage and Cloud Logging are normally enabled by default. If your organization disables these APIs, you will need to enable them manually._
 
+### Permissions required for the deploying user
+
+`roles/owner` on the target project is sufficient and is the simplest option.
+
+If your organization's policies require narrower scopes, the deploying user needs at minimum the following predefined roles. **Do not rely on `roles/editor` as a catch-all** — Google has narrowed the basic roles, and `editor` no longer includes Firestore native-mode DB creation, App Engine app creation, custom role creation, or project-level IAM bindings.
+
+| Role | Why it's needed |
+|---|---|
+| `roles/serviceusage.serviceUsageAdmin` | Enable APIs (`gcloud services enable`) and create service identities (Vertex AI service agent) |
+| `roles/storage.admin` | Create the GCS bucket, set CORS, upload example assets |
+| `roles/datastore.owner` | Create the two Firestore native-mode databases |
+| `roles/firebase.admin` | Initialise the Firebase project, create the Web App, deploy Firestore/Storage rules |
+| `roles/appengine.appCreator` | One-time `gcloud app create` (project-level App Engine app) |
+| `roles/appengine.appAdmin` | Deploy the UI to App Engine (used by `deploy-ui.sh`) |
+| `roles/run.admin` | Deploy the backend to Cloud Run |
+| `roles/artifactregistry.admin` | Create the Docker repository for the backend image |
+| `roles/apigateway.admin` | Create the API Gateway, its config, and its managed service |
+| `roles/cloudtasks.admin` | Create the three task queues (Veo, Gemini, Other) |
+| `roles/serviceusage.apiKeysAdmin` | Create the API key used by the gateway |
+| `roles/iam.serviceAccountAdmin` | Service-account-level IAM bindings (e.g. Cloud Tasks → Cloud Run actAs) |
+| `roles/iam.serviceAccountUser` | Allow the deploy to act as service accounts during Cloud Run deploy |
+| `roles/iam.roleAdmin` | Create the custom `SceneMachineUser` role used for end-user authorization |
+| `roles/resourcemanager.projectIamAdmin` | Project-level IAM bindings (used throughout for the default Compute Engine SA and the Vertex AI service agent) |
+
+The two distinct `setIamPolicy` permissions matter: `projectIamAdmin` covers project-level bindings, `serviceAccountAdmin` covers service-account-level bindings. Both come up.
+
+End users of the deployed app only need the custom `projects/$PROJECT/roles/SceneMachineUser` role — see [Adding Users](#adding-users).
+
 ## Deployment
 
 [< Technical Requirements](#technical-requirements) • [Top](#readme-top) • [Adding Users >](#adding-users)
--- a/README.md
+++ b/README.md
@@ -97,27 +97,20 @@ _Please note that most of the APIs are enabled automatically when you run the de
 
 `roles/owner` on the target project is sufficient and is the simplest option.
 
-If your organization's policies require narrower scopes, the deploying user needs at minimum the following predefined roles. **Do not rely on `roles/editor` as a catch-all** — Google has narrowed the basic roles, and `editor` no longer includes Firestore native-mode DB creation, App Engine app creation, custom role creation, or project-level IAM bindings.
+If your organization's policies require narrower scopes, the tested minimum set is `roles/editor` plus five add-ons. `roles/editor` still covers most of the deploy work (API enable, GCS bucket, Cloud Run, Artifact Registry, API Gateway, Cloud Tasks, API keys, Firebase init, "act as" service accounts). Google narrowed the basic roles for a handful of specific operations, which is what the add-ons cover.
 
 | Role | Why it's needed |
 |---|---|
-| `roles/serviceusage.serviceUsageAdmin` | Enable APIs (`gcloud services enable`) and create service identities (Vertex AI service agent) |
-| `roles/storage.admin` | Create the GCS bucket, set CORS, upload example assets |
-| `roles/datastore.owner` | Create the two Firestore native-mode databases |
-| `roles/firebase.admin` | Initialise the Firebase project, create the Web App, deploy Firestore/Storage rules |
-| `roles/appengine.appCreator` | One-time `gcloud app create` (project-level App Engine app) |
-| `roles/appengine.appAdmin` | Deploy the UI to App Engine (used by `deploy-ui.sh`) |
-| `roles/run.admin` | Deploy the backend to Cloud Run |
-| `roles/artifactregistry.admin` | Create the Docker repository for the backend image |
-| `roles/apigateway.admin` | Create the API Gateway, its config, and its managed service |
-| `roles/cloudtasks.admin` | Create the three task queues (Veo, Gemini, Other) |
-| `roles/serviceusage.apiKeysAdmin` | Create the API key used by the gateway |
-| `roles/iam.serviceAccountAdmin` | Service-account-level IAM bindings (e.g. Cloud Tasks → Cloud Run actAs) |
-| `roles/iam.serviceAccountUser` | Allow the deploy to act as service accounts during Cloud Run deploy |
-| `roles/iam.roleAdmin` | Create the custom `SceneMachineUser` role used for end-user authorization |
-| `roles/resourcemanager.projectIamAdmin` | Project-level IAM bindings (used throughout for the default Compute Engine SA and the Vertex AI service agent) |
+| `roles/editor` | Catch-all for the bulk of the deploy operations |
+| `roles/datastore.owner` | Firestore native-mode database creation (not in `editor`) |
+| `roles/appengine.appCreator` | One-time `gcloud app create` at the project level (separate from `appAdmin`; not in `editor`) |
+| `roles/resourcemanager.projectIamAdmin` | Project-level IAM bindings — the `add_iam_binding` calls in `deploy.sh` |
+| `roles/iam.roleAdmin` | Create the custom `SceneMachineUser` role late in `deploy.sh` |
+| `roles/iam.serviceAccountAdmin` | Service-account-level IAM bindings (e.g. Cloud Tasks → Cloud Run "actAs") — a separate `setIamPolicy` from the project-level one |
 
-The two distinct `setIamPolicy` permissions matter: `projectIamAdmin` covers project-level bindings, `serviceAccountAdmin` covers service-account-level bindings. Both come up.
+For `deploy-ui.sh`, add `roles/appengine.appAdmin` so the UI can be deployed to the existing App Engine app (note: this is distinct from `appCreator`, which only allows the initial app creation).
+
+Two distinct `setIamPolicy` permissions matter: `projectIamAdmin` covers project-level bindings; `serviceAccountAdmin` covers service-account-level bindings. `deploy.sh` uses both.
 
 End users of the deployed app only need the custom `projects/$PROJECT/roles/SceneMachineUser` role — see [Adding Users](#adding-users).
 
--- a/README.md
+++ b/README.md
@@ -182,20 +182,25 @@ _Please note that most of the APIs are enabled automatically when you run the de
     - Go to the [Firebase console](https://console.firebase.google.com/), select your project then click **Authentication > Sign-in method**.
     - Click **Add new provider**, choose **Google** then enable and save it.
 
-6.  **Link Storage Bucket to Firebase**
-    - Go to the [Firebase console](https://console.firebase.google.com/).
-    - Select your project from the list.
-    - Hover over **Databases & Storage** on the left menu and select **Storage**.
-    - Click **Get started** (if prompted) and ensure the location selected is the same as you selected for the project.
-      - _Note: No-cost locations are only available in the USA._
-    - If asked, start in production mode to ensure the data in firebase is kept private.
-    - Click on the bucket name dropdown and select **+ Add bucket** (if not automatically prompted).
-    - Select **Import existing Google Cloud Storage buckets**.
-    - Select your project bucket and confirm.
+6.  **Link Storage Bucket to Firebase** — _two sequential actions_ are required on the same Firebase Storage page.
+    - Open the [Firebase console](https://console.firebase.google.com/) → select your project → **Databases & Storage** → **Storage**.
+
+    **(a) Initialize Firebase Storage** (one-time per project):
+    - Click **Get started** and walk through the wizard. This creates the project's *default* `<project>.firebasestorage.app` bucket, which the Firebase CLI requires; without it the storage deploy fails with `Firebase Storage has not been set up on project`.
+    - Production mode is fine.
+    - _Note: No-cost locations are only available in the USA._
+
+    **(b) Register your project bucket with Firebase Storage:**
+    - After (a) finishes the bucket dropdown appears at the top of the page (it doesn't exist until a bucket exists).
+    - Click the dropdown → **+ Add bucket** → **Import existing Google Cloud Storage buckets**.
+    - Select your project bucket (the one referenced by `GCS_BUCKET` in `config.txt`) → confirm.
 
 7.  **Deploy UI**
     - Run `./deploy-ui.sh` (if you skipped it during backend deployment).
     - If requested, perform any required manual steps indicated by the script (e.g. linking buckets or configuring OAuth).
+
+> [!TIP]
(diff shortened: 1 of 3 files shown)


christophervoelpel:
@victor-paunescu @adamread I've been deploying SM 4 times in fresh (non-corp GCP).
I've summarized the issues I've phased and proposed a couple of improvements to the deploy scripts (BE, UI).
I'd like to get your feedback on the mechanisms (like tool check, check gcloud project vs. config project - happened to me) as well as the new gates instead of stopping the script and then general formatting (using [>] to outline steps in the console printout rather for easier identification linking back to the script).

deploy.sh
```diff
@@ -53,13 +152,88 @@ if [ $MISSING -gt 0 ]; then
...
+  echo "  ⚠ Project mismatch — this script will switch gcloud's active project"
+  echo "    to '$PROJECT' before deploying."
+fi
+echo "════════════════════════════════════════════════════════════════════════"
+if [ ! -t 0 ]; then
+  echo "ERROR: stdin is not a TTY — cannot confirm. Re-run interactively."
+  exit 1
+fi
+read -r -p "Proceed and deploy to '$PROJECT'? (y/N) " confirm
+case "$confirm" in
```
adamread: The script specifies bash, so you could use variable expansion here to make the check a bit cleaner.

The change would be something like:
```
read -r -p "Proceed... (y/N)" confirm
# The ,, operator converts the variable's value to lowercase
confirm="${confirm,,}"

if ! [[ "$confirm" == "yes" || "$confirm" == "y" ]]; then
  echo "Aborted..."
  exit 1
fi
```

deploy.sh
```diff
@@ -13,15 +13,114 @@
...
+    local backoff=$((base + jitter))
+    [ $backoff -lt 1 ] && backoff=1
+
+    echo "  IAM binding${label} hit a transient conflict — retrying in ${backoff}s (attempt $attempt/$max_attempts)..."
+    sleep $backoff
+    attempt=$((attempt + 1))
+  done
+}
+
 set -eu
```
adamread: Please change to `set -euo pipefail` to ensure errors in pipes are also fatal.

deploy.sh
```diff
@@ -102,14 +278,17 @@ envsubst < ./firebase/firebase.template.json > ./firebase/firebase.json
...
 
-echo "Deploying rules for Backend Firestore DB..."
+echo
+echo "[>] Deploying rules for Backend Firestore DB..."
 firebase deploy --config firebase/firebase.json --only firestore --project $PROJECT
 
 rm firebase/firebase.json
 rm firebase/.firebaserc
 
 export FIREBASE_API_KEY=$(firebase --non-interactive --project $PROJECT apps:sdkconfig WEB | grep '"apiKey":' | awk -F '"' '{print $4}')
```
adamread: This could silently fail. shellscheck will suggest separating assignment and declaration, but the set at the top should also include `-o pipefail` to ensure the script dies if there's an error in this pipeline (added as a separate comment).

deploy.sh
```diff
@@ -141,11 +337,27 @@ ROLES=(
...
+echo "✓ Roles granted."
+
+# Explicitly grant the Vertex AI service agent storage access. The agent is
+# auto-granted this on first use but propagation lags by ~5-15 min; binding it
+# now eliminates the first-Veo-generation "Service agents are being
+# provisioned" failure (see DEPLOY_NOTES.md issue #6).
+AIPLATFORM_SA="service-${PROJECT_NUMBER}@gcp-sa-aiplatform.iam.gserviceaccount.com"
+echo
+echo "[>] Granting roles/storage.objectUser to Vertex AI service agent..."
+add_iam_binding $PROJECT \
```
adamread: The use of the continuation is inconsistent. Google's style guide says lines shouldn't be longer than 80 characters, except when they aren't. I'm personally indiferent, but we should be consistent.

deploy.sh
```diff
@@ -262,13 +480,53 @@ else
...
+echo "           been set up on project'."
+echo "       (b) On the same page, AFTER (a) finishes (the bucket dropdown"
+echo "           only appears once a bucket exists), click the dropdown →"
+echo "           '+ Add bucket' → 'Import existing Google Cloud Storage"
+echo "           buckets' → select ${GCS_BUCKET} → confirm. Registers your"
+echo "           project bucket so deploy-ui.sh can target it."
+echo
+echo "════════════════════════════════════════════════════════════════════════"
+echo
+read -r -p "Run ./deploy-ui.sh now? (y/N) " answer
```
adamread, line 526: This is inconsistent with getting the answer earlier. This should likely also accept yes.

README.md
adamread, line 1: Please format the document.

README.md
```diff
@@ -92,6 +93,27 @@ The following APIs are used by Scene Machine:
...
+### Permissions required for the deploying user
+
+`roles/owner` on the target project is sufficient and is the simplest option.
+
+If your organization's policies require narrower scopes, the tested minimum set is `roles/editor` plus five add-ons. `roles/editor` still covers most of the deploy work (API enable, GCS bucket, Cloud Run, Artifact Registry, API Gateway, Cloud Tasks, API keys, Firebase init, "act as" service accounts). Google narrowed the basic roles for a handful of specific operations, which is what the add-ons cover.
+
+| Role | Why it's needed |
+|---|---|
+| `roles/editor` | Catch-all for the bulk of the deploy operations |
+| `roles/datastore.owner` | Firestore native-mode database creation (not in `editor`) |
```
adamread: I'm not sure the "not in editor" comments are necessary. They're only here and on the next line, and I think the assumption can be made that they aren't, as they wouldn't be listed if they were.

README.md
```diff
@@ -92,6 +93,27 @@ The following APIs are used by Scene Machine:
...
+| `roles/appengine.appCreator` | One-time `gcloud app create` at the project level (separate from `appAdmin`; not in `editor`) |
+| `roles/resourcemanager.projectIamAdmin` | Project-level IAM bindings — the `add_iam_binding` calls in `deploy.sh` |
+| `roles/iam.roleAdmin` | Create the custom `SceneMachineUser` role late in `deploy.sh` |
+| `roles/iam.serviceAccountAdmin` | Service-account-level IAM bindings (e.g. Cloud Tasks → Cloud Run "actAs") — a separate `setIamPolicy` from the project-level one |
+
+For `deploy-ui.sh`, add `roles/appengine.appAdmin` so the UI can be deployed to the existing App Engine app (note: this is distinct from `appCreator`, which only allows the initial app creation).
```
adamread: I think this can be included in the table.

README.md
```diff
@@ -92,6 +93,27 @@ The following APIs are used by Scene Machine:
 
 _Please note that most of the APIs are enabled automatically when you run the deployment script. Cloud Storage and Cloud Logging are normally enabled by default. If your organization disables these APIs, you will need to enable them manually._
 
+### Permissions required for the deploying user
+
+`roles/owner` on the target project is sufficient and is the simplest option.
+
+If your organization's policies require narrower scopes, the tested minimum set is `roles/editor` plus five add-ons. `roles/editor` still covers most of the deploy work (API enable, GCS bucket, Cloud Run, Artifact Registry, API Gateway, Cloud Tasks, API keys, Firebase init, "act as" service accounts). Google narrowed the basic roles for a handful of specific operations, which is what the add-ons cover.
```
adamread: > Google narrowed the basic roles for a handful of specific operations, which is what the add-ons cover.

Maybe, "The following additional roles are not included in `editor`, but are required to deploy Scene Machine."

README.md
```diff
@@ -92,6 +93,27 @@ The following APIs are used by Scene Machine:
...
+| `roles/editor` | Catch-all for the bulk of the deploy operations |
+| `roles/datastore.owner` | Firestore native-mode database creation (not in `editor`) |
+| `roles/appengine.appCreator` | One-time `gcloud app create` at the project level (separate from `appAdmin`; not in `editor`) |
+| `roles/resourcemanager.projectIamAdmin` | Project-level IAM bindings — the `add_iam_binding` calls in `deploy.sh` |
+| `roles/iam.roleAdmin` | Create the custom `SceneMachineUser` role late in `deploy.sh` |
+| `roles/iam.serviceAccountAdmin` | Service-account-level IAM bindings (e.g. Cloud Tasks → Cloud Run "actAs") — a separate `setIamPolicy` from the project-level one |
+
+For `deploy-ui.sh`, add `roles/appengine.appAdmin` so the UI can be deployed to the existing App Engine app (note: this is distinct from `appCreator`, which only allows the initial app creation).
+
+Two distinct `setIamPolicy` permissions matter: `projectIamAdmin` covers project-level bindings; `serviceAccountAdmin` covers service-account-level bindings. `deploy.sh` uses both.
```
adamread: I'd leave this out. All of the roles matter.

adamread (changes requested):
Please run shellcheck on the script files. I think the only errors you'll get are for quoting variables and separating declaration and assignment. The tool will show you the lines though and I think your AI helper of choice could likely use the output.

Also, error output should go to stderr. This is done in the `add_iam_binding` function, but I don't think it's done anywhere else.

deploy.sh
christophervoelpel: Done in 6f79ce1

deploy.sh
christophervoelpel: moved to one-line for all add_iam_binding instances

deploy.sh
christophervoelpel: Accepts yes.

Not sure what "This is inconsistent with getting the answer earlier. " means.

christophervoelpel:
> Please run shellcheck on the script files. I think the only errors you'll get are for quoting variables and separating declaration and assignment. The tool will show you the lines though and I think your AI helper of choice could likely use the output.
> 
> Also, error output should go to stderr. This is done in the `add_iam_binding` function, but I don't think it's done anywhere else.

Applied shellcheck - summarized output below.
I would defer to a later change as it's mostly quoting issues which were already present.

<h3 style="margin-top: 0px; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(31, 31, 31); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Shellcheck results</h3>

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(31, 31, 31); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Exactly what adamread predicted. No surprises. The findings cluster on:</p><ul style="padding-inline-start: 2em; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(31, 31, 31); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li><strong>SC2086</strong><span> </span>(83 sites): bare<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">$PROJECT</code><span> </span>/<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">$REGION</code><span> </span>/<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">$GCS_BUCKET</code><span> </span>etc. in<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">gcloud</code><span> </span>args. Fix is mechanical: add quotes (<code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">"$PROJECT"</code>).</li><li><strong>SC2155</strong><span> </span>(8 sites):<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">export VAR=$(cmd)</code><span> </span>masks<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">cmd</code>'s exit code. Fix: split into<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">VAR=$(cmd)</code><span> </span>+<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">export VAR</code>.</li><li><strong>SC1091</strong><span> </span>(2 sites):<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">source ./config.txt</code><span> </span>— shellcheck can't statically follow the path. Informational only; either pass<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">-x</code><span> </span>to shellcheck or add a<span> </span><code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;"># shellcheck source=/dev/null</code><span> </span>directive above the line.</li></ul>

deploy.sh
adamread: True. Sorry. However, this also accepts "Yeah...no." 

I was referring to the read on line 188 and the related check. There, only "y" or "yes" are accepted. 

If you just want to get a "y" or anything else, then you could use `-n`, but that might be confusing too.

deploy.sh
christophervoelpel: I would defer to a later change as I was not touching this line originally

README.md
christophervoelpel: done

adamread: nit: it doesn't seem like mdformat was run on the file. The Google version will wrap everything to 80 columns.

deploy.sh
```diff
@@ -13,15 +13,114 @@
...
+echo
+echo "================================================================================"
+echo "  Scene Machine backend deploy — running pre-flight checks first..."
+echo "================================================================================"
+# --- Pre-flight: required tools and gcloud auth -----------------------------
+# Fail fast if a required command is missing or gcloud isn't authenticated,
+# rather than 30+ seconds into a gcloud/firebase call with a confusing error.
+echo "[>] Checking required tools..."
+MISSING_TOOLS=0
+require_tool() {
```
adamread, line 95: nit: I missed this, but functions should normally be defined at the top of the file, not mixed in with the rest of the script.

deploy.sh
adamread: nit: I'm still not happy that there are two ways of doing the same thing.

adamread (approved):
Some nits, but otherwise LGTM.

PR #89, opened 2026-05-12
Fix for the outpainter and converter: on save if not RGB, then convert to RGB

akushnarov:


workflow_examples/outpaint_horizontal_to_vertical_image.json
victor-paunescu, line 1: nit in the name: outpaint_**horizontal**_to_vertical_image

actions/outpaint_image.py
```diff
@@ -96,6 +96,8 @@ def _crop_image(
 
 def _image_to_bytes(img: PIL.Image.Image, image_format: str = 'PNG') -> bytes:
   """Converts a PIL Image to bytes."""
+  if img.mode not in ('RGB', 'L'):
```
victor-paunescu, line 99: Perhaps we should test this with an 'L' image (I think it's greyscale, right?), just to make sure we don't need to also convert those to RGB.

victor-paunescu (dismissed):

actions/outpaint_image.py
```diff
@@ -120,6 +122,8 @@ def execute(
     image: the relative path of the image to outpaint.
     target_ratio: the aspect ratio of the finished image in the format
       width:height (e.g. 9:16)
+    outpainter_model: the model to use for outpainting
+    outpainter_model_location: the location of the model to use for outpainting
```
AndreyMarkinPPC: Nit here and above. Capitalize after colon -> `outpainter_model: The model ...`.
The sentence should end with a period.

AndreyMarkinPPC (approved):

actions/outpaint_image.py
akushnarov: Added a test

actions/outpaint_image.py
akushnarov: Thanks!

akushnarov (commented):
Fixed

victor-paunescu (approved):

PR #104, opened 2026-06-04
Upgrade to GA Gemini 3 models and improve outpainting

christophervoelpel:
## Summary

Two related changes (the outpainting improvement depends on the model/config update):

### 1. Upgrade models & unify image config
- Text generation → `gemini-3.5-flash`
- Image generation / outpainting → `gemini-3-pro-image` (Nano Banana Pro)
- Veo default → `veo-3.1-generate-001`
- Consolidate the outpainting and image-generation model settings under a single
  `IMAGE_MODEL` / `IMAGE_REGION` (replacing `OUTPAINTER_MODEL` / `OUTPAINTER_REGION`),
  threaded through the actions, UI services, deploy scripts, README, and workflow examples.

### 2. Improve outpainting (extend, don't re-imagine)
Transforms an image to a target aspect ratio without the model regenerating the scene as a
side-by-side composite:
- Send a blank target-ratio canvas as a second image, so the model adopts the target ratio
  and fills only the new margins.
- A stronger "extend the existing scene, do not regenerate" prompt.
- Output resolution adapts (2K, stepping up to 4K for large target canvases) so results
  aren't downscaled on big expansions.
- Request a HIGH thinking level on models that advertise thinking support (queried from the
  model registry; fails closed so non-thinking models don't error).

## Testing
- `python3 -m pytest actions_lib/test/test_outpainter.py` → 9 passed
- `python3 -m pytest actions_lib/test/test_gemini.py` → 10 passed


actions_lib/image_creator.py
```diff
@@ -20,13 +20,11 @@
...
 
 
-IMAGE_MODEL = "gemini-2.5-flash-image"
-
-
 def generate_images(
     gcp_project: str,
     gcp_location: str,
     image_prompt: str,
+    image_model: str,
```
victor-paunescu, line 30: Do we also need image_model_location here?

actions_lib/outpainter.py
```diff
@@ -30,60 +36,171 @@
...
+    skip is logged so it is observable rather than silent.
+    """
+    if model not in _THINKING_SUPPORT:
+        try:
+            _THINKING_SUPPORT[model] = bool(client.models.get(model=model).thinking)
+        except Exception:  # pylint: disable=broad-except
+            logger.warning("Could not query thinking support for %s", model)
+            _THINKING_SUPPORT[model] = False
+        if not _THINKING_SUPPORT[model]:
+            logger.info("Thinking not enabled for model %s", model)
```
victor-paunescu: Is it an issue if the model doesn't have thinking support?

actions_lib/outpainter.py
```diff
@@ -30,60 +36,171 @@
...
+    return _THINKING_SUPPORT[model]
+
 
 def outpaint_image(
     image_bytes: bytes,
     gcp_project: str,
     gcp_location: str,
-    outpainter_model: str,
+    image_model: str,
```
victor-paunescu, line 152: Do we also need image_model_location here?

victor-paunescu (changes requested):
I think we also need to update the definitions file
`ui/definitions/actions.json`

actions_lib/outpainter.py
christophervoelpel: We could - it was not mentioned here previously and lind 183 uses gcp_location for location.
WDYT - blocking or nit ?

actions_lib/outpainter.py
christophervoelpel: Line 194 conditionally hands over thinking as part of the config_kwargs for the NB request. It looked like the cleaner version to only append the config_kwargs if thinking is supported by the model to be called.

WDYT blocking / nit?

christophervoelpel (commented):
added comments on the comments - asking for your assessment whether nit or blocking

actions_lib/outpainter.py
victor-paunescu: Actually the `image_model_location` is being passed as `gcp_location` here, so it should be fine.

actions_lib/outpainter.py
victor-paunescu: That's fine. My comment was more about why we would log that thinking is not enabled for the model.

actions/generate_image.py
victor-paunescu: Perhaps it would make sense to also add an `image_model_location` here and pass it to `gcp_location` like we do for the outpainter.

actions_lib/image_creator.py
victor-paunescu: I added a suggestion in a comment above for this.

actions/generate_image.py
```diff
@@ -33,7 +33,8 @@ def execute(
     workflow_params: Params,
     prompt: NodeInput,
     variant_quantity: int,
-    aspect_ratio: str,
+    image_model: str,
+    aspect_ratio: str = '16:9',
```
akushnarov: Let's not use default values in actions, the workflow creator should define the aspect ratio

actions/generate_image.py
akushnarov: We already have OUTPAINTER_REGION in config, maybe we can rename it to the IMAGE_MODEL_REGION

actions/translate.py
akushnarov, line 1: Let's remove this file, it's just an example

actions_lib/test/test_image_describer.py
```diff
@@ -32,14 +32,14 @@ def test_describe_image_with_guidance(self, mock_prompt):
         image_path='gs://bucket/image.png',
         guidance='a red car',
         gcp_project='test-proj',
-        gemini_model='gemini-2.5-flash',
+        gemini_model='gemini-3.5-flash',
```
akushnarov: Let's use model names like 'mock-model' since here we don't care about the exact model name.

actions/test/test_write_products_script.py
```diff
@@ -173,7 +173,7 @@ def test_execute_raises_runtime_error(self, mock_prompt):
           product_description=product_description,
           image_description=image_description,
           story_variant_quantity=1,
-          gemini_model="gemini-2.5-flash",
+          gemini_model="gemini-3.5-flash",
```
akushnarov: Let's use model names like 'mock-model' since here we don't care about the exact model name.

actions/test/test_analyse_file.py
```diff
@@ -44,7 +44,7 @@ def test_execute_simple(self, mock_gemini_prompt):
         file=file_input,
         generation_prompt=None,
         response_schema={},
-        gemini_model="gemini-2.5-flash",
+        gemini_model="gemini-3.5-flash",
```
akushnarov: Let's use model names like 'mock-model' since here we don't care about the exact model name.

akushnarov (changes requested):

ui/definitions/actions.json
```diff
@@ -131,6 +131,12 @@
     "parameters": {
       "target_ratio": {
         "type": "string"
+      },
```
christophervoelpel, line 134: @victor-paunescu 
Added image_model and image_model_location to both the generate_image and outpaint_image definitions so they match the action signatures .

actions/generate_image.py
```diff
@@ -55,11 +60,12 @@ def execute(
   # generate_images now returns a list of (image_bytes, mime_type) tuples
   generated_data = image_creator.generate_images(
```
christophervoelpel, line 61: @victor-paunescu Re: "need image_model_location here?"
Yes. generate_image now takes image_model_location and passes it as the model location, the same way the outpainter does.

actions/generate_image.py
```diff
@@ -34,7 +34,8 @@ def execute(
     prompt: NodeInput,
     variant_quantity: int,
     image_model: str,
-    aspect_ratio: str = '16:9',
+    image_model_location: str,
+    aspect_ratio: str,
```
christophervoelpel, line 38: @akushnarov Removed the aspect_ratio = '16:9' default. The workflow creator sets it now.

actions/translate.py
christophervoelpel: I would like to keep translate.py in this PR for now. It is used by workflow_examples/demo.json, and DEVELOPING.md points at that demo as a way to exercise Remix Engine, so removing it cleanly also means deleting demo.json and updating the developer guide, which is a bit beyond the scope of this change. I have a larger cleanup PR coming and will remove translate.py and tidy up the demo and docs there.

actions/test/test_outpaint_image.py
```diff
@@ -240,6 +240,22 @@ def test_execute_invalid_ratio_raises_error(self):
           'test-location',
       )
 
+  def test_execute_degenerate_ratio_raises_error(self):
+    """Tests that zero, negative, and non-integer ratios are rejected."""
+    degenerate_ratios = ['16:0', '0:9', '0:0', '16:-9', '16:nine']
```
christophervoelpel, line 245: @victor-paunescu @akushnarov added a test to catch ratios we would fail with normally (ie. 16:0 would cause division by zero)

actions_lib/gemini.py
```diff
@@ -68,9 +68,7 @@ def _get_thinking_config(model: str):
     """Gets the thinking configuration for the specific model."""
     if model.startswith("gemini-2.5-pro"):
         return types.ThinkingConfig(thinking_budget=128)
-    elif model.startswith("gemini-2.5-flash") or model.startswith(
-        "gemini-2.5-flash-lite"
```
christophervoelpel, line 72: Removed because startswith ...-flash already catches and the or is redundant for -flash-lite

actions_lib/image_creator.py
```diff
@@ -19,6 +19,9 @@
 from google import genai
 from google.genai import types as gtypes
 
+from common import get_api_client_headers
```
christophervoelpel, line 22: @victor-paunescu @akushnarov  - new (see the pr text above)

actions_lib/test/test_outpainter.py
```diff
@@ -140,6 +140,35 @@ def test_thinking_skipped_when_unsupported(
         _, kwargs = mock_client.models.generate_content.call_args
         self.assertIsNone(kwargs["config"].thinking_config)
 
+    def test_supports_thinking_failure_not_cached(self):
```
christophervoelpel, line 143: additional finding from claudes ultra code review

actions_lib/outpainter.py
```diff
@@ -129,17 +129,14 @@ def _supports_thinking(client: genai.Client, model: str) -> bool:
...
+    a model without thinking can never hard-error on the thinking config.
     """
     if model not in _THINKING_SUPPORT:
         try:
             _THINKING_SUPPORT[model] = bool(client.models.get(model=model).thinking)
         except Exception:  # pylint: disable=broad-except
             logger.warning("Could not query thinking support for %s", model)
             _THINKING_SUPPORT[model] = False
-        if not _THINKING_SUPPORT[model]:
-            logger.info("Thinking not enabled for model %s", model)
```
christophervoelpel, line 142: @victor-paunescu @akushnarov removed thinking logger - but kept the log above

actions/test/test_analyse_file.py
```diff
@@ -44,7 +44,7 @@ def test_execute_simple(self, mock_gemini_prompt):
         file=file_input,
         generation_prompt=None,
         response_schema={},
-        gemini_model="gemini-3.5-flash",
+        gemini_model="mock-model",
```
christophervoelpel, line 47: @akushnarov per your comment about mock-model

christophervoelpel (commented):
currently deploying to gcp and corp

victor-paunescu (approved):

akushnarov (approved):

actions_lib/gemini.py
```diff
@@ -68,10 +68,12 @@ def _get_thinking_config(model: str):
...
-    ):
+    elif model.startswith("gemini-2.5-flash"):  # includes gemini-2.5-flash-lite
```
AndreyMarkinPPC, line 71: nit: Since you return on line 70, you can use `if` here. Same is applied to line 73

```
if model.startswith("..."):
  return ...
if model.startswit("..."):
  return ...
return
```

actions_lib/outpainter.py
```diff
@@ -30,60 +36,173 @@
...
     target_ratio: str,
 ) -> Tuple[bytes, str]:
-    """Outpaints the given image using Google Gemini Image.
+    """Outpaints the given image to the target aspect ratio using Gemini.
 
     Args:
-      image_bytes: the bytes of the image to outpaint. These must be a JPEG or PNG
-        image.
-      gcp_project: the ID of Google Cloud project to use with Imagen.
+      image_bytes: the bytes of the image to outpaint. Must be a JPEG or PNG.
```
AndreyMarkinPPC, line 158: Nit:  First letter should be capitalized. here and below in the docstring.

AndreyMarkinPPC (approved):

PR #106, opened 2026-06-05
feat(deploy): unify backend and UI deployment under deploy.sh - updated README

christophervoelpel:
1. Refactored argument parsing and added targeted deployment flags ( --backend-only ,  --ui-only ).                                                                                                                    
  2. Optimized API enablement with an API cache comparison to speed up subsequent runs.                                                                                                                                  
  3. Made pre-flight tool requirements component-aware so they only check for what is actually needed.                                                                                                                   
  4. Deprecated  deploy-ui.sh  and redirected all UI flows through the unified  deploy.sh .                                                                                                                              
  5. Added startup pre-flight checks to ensure the Firebase CLI session is actively authenticated.                                                                                                                       
  6. Optimized GCS/Firebase linkage checks to avoid timeouts on accounts with 1,500+ active projects.                                                                                                                    
  7. Restored direct Firestore API curl loops for database configurations upload.                                                                                                                                        
  8. Added environment override capability for  CLOUD_RUN_URL .                                                                                                                                                          
  9. Enforced OAuth Consent Screen setup via a blocking 15-second polling loop.                                                                                                                                          
  10. Enforced App Engine IAP configuration via a blocking 15-second polling loop.


deploy-ui.sh
```diff
@@ -13,390 +13,5 @@
...
-echo "       gcloud projects add-iam-policy-binding ${PROJECT} \\"
-echo "         --member=\"user:YOUR_EMAIL@example.com\" \\"
-echo "         --role=\"projects/${PROJECT}/roles/SceneMachineUser\""
-echo
-echo "       Or via the IAM console:"
-echo "       https://console.cloud.google.com/iam-admin/iam?project=${PROJECT}"
-echo
-echo "════════════════════════════════════════════════════════════════════════"
-echo
+echo "WARNING: deploy-ui.sh is deprecated. Redirecting to './deploy.sh --ui-only'..."
```
akushnarov: Let's remove deploy-ui.sh, no need to support it in the future

deploy.sh
```diff
@@ -501,46 +616,326 @@ echo "════════════════════════
...
+  echo "  Before doing that, the following manual console steps are required:"
+fi
 echo
 echo "    1. Configure OAuth consent screen"
 echo "       https://console.cloud.google.com/auth/branding?project=${PROJECT}"
 echo "       First time on this project: click 'Get started' and walk through"
-echo "       the setup dialog. User Type is set under 'Audience' — pick"
-echo "       'Internal' if you have a Workspace org; otherwise 'External' and"
-echo "       add yourself as a test user."
+echo "       the setup dialog. Set User Type to 'Internal' for @google.com logins."
```
akushnarov: Why `@google.com`?
Proposed: "In the OAuth setup dialog, set User Type to 'Internal' if you only want users within *your own* Google Workspace organization to access the app (recommended for testing), or 'External' if you want anyone to be able to log in."

README.md
```diff
@@ -329,17 +329,40 @@ End users of the deployed app only need the custom
...
 
+#### Target Options and Non-Interactive (Agent) Deployment
+
+By default, `./deploy.sh` will deploy both the backend and UI. You can use targeting flags to deploy them individually:
+
+* **Deploy Backend Only**: `./deploy.sh --backend-only` (or `--be-only`)
+* **Deploy UI Only**: `./deploy.sh --ui-only`
+* **Local Development Build**: `./deploy.sh --ui-only --local` (or `./deploy.sh --ui-only local`) to build the Angular application files locally without triggering an App Engine deployment.
+
+To run the script in a non-interactive/headless environment (such as an AI agent like Jetski or Claude), use the `-y` / `--yes` / `--non-interactive` flag to auto-confirm target configuration prompts:
```
akushnarov: a bit confusing to have too many flags with the same functionality, I would keep only standard `-y` & `--yes`

deploy.sh
```diff
@@ -104,6 +104,48 @@ require_tool() {
...
+TARGETED=false
+DEPLOY_MODE=""
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    -y|--yes|--non-interactive)
+      AUTO_CONFIRM=true
+      shift
+      ;;
+    --backend-only|--be-only)
```
akushnarov: let's leave only one `--be-only`

deploy.sh
```diff
@@ -104,6 +104,48 @@ require_tool() {
...
 set -euo pipefail
+
+# Parse arguments
+AUTO_CONFIRM=false
+DEPLOY_BACKEND=false
+DEPLOY_UI=false
+TARGETED=false
+DEPLOY_MODE=""
+
+while [[ $# -gt 0 ]]; do
```
akushnarov, line 39: Let's have flags `-h` & `--help` for help (to output all the available options)?

akushnarov (changes requested):

deploy-ui.sh
adamread: +1. This seems more confusing than helpful to new users.

deploy.sh
adamread: +1. I'd say `--backend-only` is better, as it's more descriptive and clear.

deploy.sh
adamread, line 1: Google tools generally use lower_snake_case for flags. Should we do this for consistency?

deploy.sh
```diff
@@ -104,6 +104,48 @@ require_tool() {
...
+  case "$1" in
+    -y|--yes|--non-interactive)
```
adamread: I think just having `--non_interactive` would suffice, as it's more descriptive and clear.

deploy.sh
```diff
@@ -104,6 +104,48 @@ require_tool() {
...
+      DEPLOY_BACKEND=true
+      TARGETED=true
+      shift
+      ;;
+    --ui-only)
+      DEPLOY_UI=true
+      TARGETED=true
+      shift
+      ;;
+    local|--local)
```
adamread, line 56: I would stick to just the flag version (i.e. `--local`) for consistency.

deploy.sh
```diff
@@ -104,6 +104,48 @@ require_tool() {
...
+      DEPLOY_MODE="local"
+      shift
+      ;;
+    *)
+      echo "Unknown argument: $1" >&2
+      echo "Usage: $0 [options]" >&2
```
adamread, line 62: I don't know if a help flag is necessary, but something more descriptive here might help.

deploy.sh
```diff
@@ -114,9 +156,13 @@ echo "==========================================================================
 echo "[>] Checking required tools..."
 require_tool gcloud   "Install: https://cloud.google.com/sdk/docs/install"
 require_tool firebase "Install: npm i -g firebase-tools"
-require_tool node     "Install Node.js ≥ v22: https://nodejs.org/en/download"
-require_tool npm      "Install Node.js (includes npm): https://nodejs.org/en/download"
 require_tool envsubst "Install gettext (macOS: 'brew install gettext'; Debian/Ubuntu: 'apt-get install gettext')"
+
+if [ "$DEPLOY_UI" = "true" ]; then
+  require_tool node     "Install Node.js ≥ v22: https://nodejs.org/en/download"
+  require_tool npm      "Install Node.js (includes npm): https://nodejs.org/en/download"
```
adamread, line 88: If node includes npm, do we need to check both?

deploy.sh
```diff
@@ -268,11 +369,17 @@ fi
 
 echo
 echo "[>] Setting up Firebase project and Web App..."
-if gcloud services list --enabled --project=$PROJECT --filter="name:firebase.googleapis.com" | grep -q "firebase.googleapis.com"; then
-  echo "Firebase is already enabled for the project."
-else
-  echo "Enabling Firebase..."
+if ! is_service_enabled "firebase.googleapis.com"; then
+  echo "Enabling Firebase Management API..."
   gcloud services enable firebase.googleapis.com --project=$PROJECT
```
adamread, line 300: Shouldn't this be with the other required APIs?

deploy.sh
```diff
@@ -268,11 +369,17 @@ fi
...
+  # Refresh cache
+  ENABLED_APIS=$(gcloud services list --enabled --project="$PROJECT" --format="value(config.name)" 2>/dev/null || true)
```
adamread, line 302: Why?

deploy.sh
adamread: I'd just say that internal is recommended for security purposes when possible.

deploy.sh
```diff
@@ -501,46 +616,326 @@ echo "════════════════════════
...
+
+  generate_config
+  envsubst < ./ui/src/env.template.txt > ./ui/src/env.ts
+
+  if is_service_enabled "identitytoolkit.googleapis.com"; then
+    echo "Identity Toolkit API is already enabled."
+  else
+    echo "Enabling Identity Toolkit API (needed for Auth config)..."
+    gcloud services enable identitytoolkit.googleapis.com --project=$PROJECT
+    ENABLED_APIS=$(gcloud services list --enabled --project="$PROJECT" --format="value(config.name)" 2>/dev/null || true)
```
adamread, line 643: also not required. It's not used again in the script.

deploy.sh
```diff
@@ -501,46 +616,326 @@ echo "════════════════════════
...
+  fi
+
+  # --- Bucket linked to Firebase Storage: poll until satisfied ----------------
+  echo
+  echo "[>] Checking if bucket ${GCS_BUCKET} is linked to Firebase Storage..."
+  BUCKET_WAIT_ATTEMPTS=0
+  BUCKET_WAIT_MAX=120   # 120 × 15s = 30 min total
+  while true; do
```
adamread, line 651: I'm not sure about this (and the polling further down) from the automated deployment perspective. If this isn't being run in an interactive session, someone may eventually see this in the logs after the deployment fails. 

I think fast fail when `AUTO_CONFIRM` is true might be a good idea. That way we don't block CI pipelines for 30 min if someone messed up the initial deployment.

We should also add a note in the readme that an automated deployment cannot be done for the first time.

README.md
adamread: as above, I would go with `--non-interactive`, as it's the most clear.

adamread (commented):
I have some comments on the flags and some minor concerns about the polling during a non-interactive deployment.

PR #113, opened 2026-06-17
Add Cloud Tasks lock to orchestrator to prevent duplicate execution

victor-paunescu:


util/database.py
```diff
@@ -361,6 +361,63 @@ def store_output(
...
+    Args:
+      execution_id: The workflow's execution ID.
+      node_id: The node ID.
+      group_id: The group ID.
+
+    Returns:
+      True if the lock was successfully acquired (first to trigger),
+      False otherwise.
+    """
+    doc_id = f'{execution_id}_{node_id}_{group_id}'
```
akushnarov: move to the function, since you can reuse it for the release_lock...

akushnarov (dismissed):

util/database.py
victor-paunescu: Done

util/database.py
```diff
@@ -361,6 +361,80 @@ def store_output(
...
+      group_id: The group ID.
+
+    Returns:
+      True if the lock was successfully acquired (first to trigger),
+      False otherwise.
+    """
+    doc_ref = self._get_task_lock_ref(execution_id, node_id, group_id)
+    expires_at = datetime.datetime.now(
+        datetime.timezone.utc
+    ) + datetime.timedelta(hours=12)
```
akushnarov: nit: I would add a variable e.g. TASK_LOCK_EXPARATION_HOURS=12, and use the variable (similar to MAX_ATTEMPTS below)

akushnarov (dismissed):

util/database.py
victor-paunescu: Good point, done

akushnarov (approved):

PR #133, opened 2026-07-09
Validate configured models against the allowlist at deploy time

christophervoelpel:
## Summary

`deploy.sh` now validates each modality's configured model and region against the model allowlist before building the image. A typo, an un-allowlisted model, a family mismatch, or a region the model doesn't support fails the deploy with a clear message instead of surfacing later at runtime against Vertex.

## Behavior

- After sourcing `config.txt` and before the build, `deploy.sh` runs `scripts/validate_config_models.py`; a non-zero exit aborts the deploy.
- The `source ./config.txt` line is wrapped in `set -a` / `set +a` so that config vars written as a bare `VAR=` (no `export`) are exported to child processes. The required-vars check accepts a bare `VAR=`, but without `set -a` such a var is invisible to the Python check's `os.environ` and renders empty in `envsubst`.
- For each modality (`GEMINI_MODEL`/`GEMINI_REGION`, `IMAGE_MODEL`/`IMAGE_MODEL_REGION`, `VEO_MODEL`/`VEO_REGION`), the check requires the model to be present in `ui/definitions/models.json`, to have the expected family (`gemini`/`image`/`veo`), and to have its configured region listed in that model's `locations`.
- A modality whose model is unset is skipped. A model that is set but has an empty region is an error, not a skip — a region is required to reach the model.
- Validation is against the local allowlist only. It does not call Vertex, so it does not prove a listed model actually resolves there.

## Tests

- `collect_errors()` is a pure function covered by unit tests for: a valid config passing, an unknown model, a disallowed region, a wrong-family model, an unconfigured modality being skipped, and a model set without a region.
- A test parses `config.template.txt`'s actual defaults from the file (rather than hard-coding them) and asserts they validate against the shipped allowlist, so a default bumped in the template but not in the allowlist fails the test.

## Risks / Notes

- Local-only validation: a model present in the allowlist but not yet confirmed on Vertex passes this check. Confirming a model resolves against Vertex is out of scope.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -215,6 +215,7 @@ require_tool node     "Install Node.js ≥ v22: https://nodejs.org/en/download"
 require_tool npm      "Install Node.js (includes npm): https://nodejs.org/en/download"
 require_tool envsubst "Install gettext (macOS: 'brew install gettext'; Debian/Ubuntu: 'apt-get install gettext')"
 require_tool curl     "Install curl (it ships with macOS and most Linux distributions)"
+require_tool python3  "Install Python 3 (macOS: 'brew install python' or the Xcode CLT; Debian/Ubuntu: 'apt-get install python3')"
 if [ $MISSING_TOOLS -gt 0 ]; then
   echo "Please install the missing tools above, then re-run $0."
   exit 1
--- a/deploy.sh
+++ b/deploy.sh
@@ -272,7 +272,25 @@ if [ $MISSING -gt 0 ]; then
   echo "Validation failed. Please fix config.txt and try again." >&2
   exit 1
 fi
+# Export sourced config vars so Python checks and envsubst see bare VAR= lines.
+set -a
 source ./config.txt
+set +a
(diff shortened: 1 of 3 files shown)


deploy.sh
```diff
@@ -291,6 +309,15 @@ fi
 # choose.
 echo "✓ config.txt is valid. Target project: $PROJECT"
 
+# --- Check configured models against the allowlist --------------------------
+# A model/region typo, or a model/region not in ui/definitions/models.json,
+# should fail here at deploy time -- not later at runtime against Vertex.
+echo "[>] Checking configured models against the allowlist..."
+if ! python3 scripts/validate_config_models.py; then
```
victor-paunescu, line 317: not sure if there would be an issue with this being `python3` instead of `python`

victor-paunescu (dismissed):

deploy.sh
christophervoelpel: Added require_tool python3

victor-paunescu (approved):

PR #134, opened 2026-07-10
Bound fan-out, quantity, and node gcp_project at submission

christophervoelpel:
## Summary

Extends the enqueue submission validator with fan-out, quantity, node-count, input-file, and per-node `gcp_project` limits. The checks inspect the values the engine will actually run, so a value wrapped in a one-element list can no longer slip past a scalar type check.

## Behavior

The validator returns `(message, code)` for the first problem it finds, else `None`. The new checks reject:

- **Node `gcp_project`** (`GCP_PROJECT_NOT_ALLOWED`) — a node may carry `gcp_project` only if every value it expands to is an empty/blank string. The server pins the project; the UI sends `''`, which is allowed. Anything else (a real project string, a number, an object, or a list whose elements aren't blank strings) is rejected, because the executor's `gcp_project or workflow_params[...]` fallback would otherwise let the node override the pinned project.
- **List too long** (`LIST_TOO_LONG`) — any list-valued node parameter longer than `_MAX_LIST_LEN` (100).
- **Per-node fan-out** (`FAN_OUT_TOO_LARGE`) — the product of one node's list-parameter lengths exceeds `_MAX_FAN_OUT` (1000). Each list can be under the length cap while their product is not.
- **Whole-workflow fan-out** (`TOTAL_FAN_OUT_TOO_LARGE`) — fan-out composes along edges, so a chain of nodes each under the per-node cap can still multiply past a sane total. The validator tracks the running product across nodes (seeded by the largest `inputFiles` group) and bounds it with `_MAX_TOTAL_FAN_OUT` (100000).
- **Quantity too large** (`QUANTITY_TOO_LARGE`) — any parameter in `_QUANTITY_PARAMS` (`video_variant_quantity`, `variant_quantity`, `story_variant_quantity`) exceeds `_MAX_QUANTITY` (100), in scalar or list form. Booleans are skipped: a `bool` is an `int` subclass but is a flag, not a quantity.
- **Too many nodes** (`TOO_MANY_NODES`) — the workflow has more than `_MAX_NODES` (200) nodes.
- **Too many input files** (`TOO_MANY_INPUT_FILES`) — an `inputFiles` group has more than `_MAX_INPUT_FILES` (200) entries.
- **Malformed `inputFiles`** (`MALFORMED_SUBMISSION`) — `inputFiles` is present but not an object. It fails closed here rather than reaching `verify_input`, whose `.items()` would raise and turn a client error into a 500.

List-shape handling: the engine expands a list parameter into one execution per element and runs a scalar as itself (`util/workflow.py` `expand_parameters` normalizes each value with `v if isinstance(v, list) else [v]`). A guard that only checks the raw value's type is therefore fooled by a one-element list — `['other']` fails an `isinstance(str)` check but the engine unwraps it to `'other'`. A helper `_as_values()` mirrors that expansion and normalizes every value to the list the engine will run; the `gcp_project` and quantity guards check those normalized values, so scalar, one-element-list, and multi-element-list forms are all covered.

Caps are named constants chosen as conservative upper bounds — generous enough that every `workflow_example` and current UI submission still passes, tight enough to stop an absurd fan-out.

## Tests

`test/test_submission_validation.py` adds coverage for each new check and its bypasses:

- One rejection test per check: node `gcp_project`, list too long, per-node fan-out product, too many nodes, too many input files, quantity too large, whole-workflow fan-out across a three-node chain, and non-object `inputFiles`.
- List-shaped inputs that must still be rejected: `gcp_project` as `['other']`, `123`, `{...}`, and `['a','b']`; each quantity parameter as `[101]`; `inputFiles` as a list.
- Happy paths that must still pass: `gcp_project=''` and `['']`, quantities within cap in scalar and list form, and a two-node workflow whose total fan-out stays under budget.

## Risks / Notes

- Cap values are chosen bounds, not derived from a measured product limit — worth confirming against real workflow sizes before merge.
- The `gcp_project` rule depends on the executor keeping the `gcp_project or workflow_params[...]` fallback; if that resolution changes, revisit the allowed-value rule.

### DIFF
--- a/util/submission_validation.py
+++ b/util/submission_validation.py
@@ -131,15 +131,34 @@ def validate_submission(
     for key, files in input_files.items():
       if not isinstance(files, list):
         return (f'inputFiles {key!r} is not a list', 'MALFORMED_SUBMISSION')
+      dimension_keys = None
       for entry in files:
         if (not isinstance(entry, dict) or not isinstance(entry.get('file'), str)
             or not entry['file'].strip()):
           return (f'inputFiles {key!r} entries must each have a non-empty '
                   f'file', 'MALFORMED_SUBMISSION')
+        # The engine groups on the non-file fields of the first entry and
+        # indexes every entry by them (group_input._group_dictionaries), so
+        # ragged keys raise KeyError there and a non-scalar value makes an
+        # unhashable group key. Fail closed on both shapes.
+        entry_keys = frozenset(entry) - {'file'}
+        if dimension_keys is None:
+          dimension_keys = entry_keys
+        elif entry_keys != dimension_keys:
+          return (f'inputFiles {key!r} entries carry inconsistent dimension '
+                  'keys', 'MALFORMED_SUBMISSION')
+        for field in entry_keys:
+          value = entry[field]
+          if value is not None and not isinstance(value, (str, int, float, bool)):
+            return (f'inputFiles {key!r} dimension {field!r} must be a scalar',
+                    'MALFORMED_SUBMISSION')
       if len(files) > _MAX_INPUT_FILES:
         return (f'inputFiles {key!r} has too many entries '
                 f'({len(files)} > {_MAX_INPUT_FILES})', 'TOO_MANY_INPUT_FILES')
-      input_group_sizes[key] = len(files)
+      # An empty group still becomes one grouping slot in the engine
+      # (group_input returns {(): []}), so counting it as 0 would zero out
+      # every downstream multiplier and hide real fan-out from the budget.
+      input_group_sizes[key] = max(1, len(files))
 
   definition = data.get(_WORKFLOW_DEFINITION)
   if definition is None:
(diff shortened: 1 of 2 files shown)


util/submission_validation.py
```diff
@@ -34,6 +34,38 @@
...
+
+# Structural / queue-safety caps -- generous; the cost cap above is the real
+# money bound. _MAX_TOTAL_FAN_OUT bounds the total node executions the
+# orchestrator must materialize (billable or not), so a pass-node blow-up can't
+# exhaust it.
+_MAX_NODES = 200            # nodes in a single workflow
+_MAX_INPUT_FILES = 200      # entries in one inputFiles group
+_MAX_LIST_LEN = 100         # length of any list-valued node parameter
+_MAX_FAN_OUT = 1000         # product of a single node's list-parameter lengths
+_MAX_TOTAL_FAN_OUT = 100000  # total node executions across the whole workflow
```
cahlheim, line 65: just double-checking that those parameters are very generous? if not, might be better to make them more explicit so that they're easier to change?

cahlheim (approved):

victor-paunescu (approved):

util/submission_validation.py
christophervoelpel: makes sense - most of this change is theoritical boundaries for remix engine (we never even came close to any of these figures).
I've put in _MAX_NODES & _MAX_GENERATIONS at 200 to ensure we are not spawning more than 200 nodes (which could cost money).

PR #137, opened 2026-07-13
Seed config/models from the repo catalog, with a diff preview

christophervoelpel:
## Summary
Every deploy writes the model catalog (`ui/definitions/models.json`) to the `config/models` Firestore document, and previews what that write will change before the deployment-confirmation prompt. Nothing reads the document yet; runtime behavior is unchanged.

## Behavior
- During pre-flight, `deploy.sh` fetches the live `config/models` doc and prints a per-entry diff — models added/removed/changed, defaults changes, and anything that exists only live — so a console edit is replaced visibly, never silently. Any preview failure (fresh project, API not enabled, unseeded doc, a Firestore-only value type in the doc) degrades to "seeding from the repo" and never blocks the deploy.
- After the existing `config/global` seed, the deploy converts the repo catalog to the Firestore REST typed-value format and PATCHes `config/models`, failing loudly on a non-200 like the other seeds. The PATCH replaces the whole document by design: the repo is the source of truth and live edits are temporary.
- `scripts/seed_config_models.py` is stdlib-only (the deploy machine has no pip environment). Booleans are checked before integers so `true` cannot be written as an `integerValue`; Firestore-only types found in the live doc (timestamp, bytes) render as markers the diff reports as overwritten; a live-only field with a `null` value is still reported as removed.
- The dev-database seed instructions gain the same document.

## Tests
- Conversion in both directions: scalar types, bool-vs-int ordering, int64-as-string, nested maps/lists, dotted model-id map keys, empty array/map encodings, Firestore-only-type markers; a round-trip assertion over the real `models.json`.
- Diff output: identical docs, unreadable/malformed live docs (degrade, exit 0), live docs with Firestore-only types (exit 0, reported as removed), live-only entries including explicit nulls, one-sided sections, and per-field change lines.

## Risks / Notes
- The pre-flight preview runs two network calls (token + GET) earlier in the deploy than any existing Firestore access; both degrade to "nothing to diff" on failure.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -58,6 +58,15 @@ curl -s -X PATCH \
   -H "x-goog-user-project: ${PROJECT}" \
   -H "Content-Type: application/json" \
   -d @<(envsubst < ./firestore_config_frontdoor.template.json)
+
+# The model catalog document (config/models), seeded from the repo file the
+# same way deploy.sh does it:
+python3 scripts/seed_config_models.py convert < ui/definitions/models.json | curl -s -X PATCH \
+  "https://firestore.googleapis.com/v1/projects/${PROJECT}/databases/${FIRESTORE_DB_UI}/documents/config/models" \
+  -H "Authorization: Bearer $(gcloud auth application-default print-access-token)" \
+  -H "x-goog-user-project: ${PROJECT}" \
+  -H "Content-Type: application/json" \
+  -d @-
 ```
 
 This requires a Firestore database to already exist in your dev project (the one `FIRESTORE_DB_UI` names) and ADC with write access to it.
--- a/deploy.sh
+++ b/deploy.sh
@@ -319,6 +319,26 @@ if ! python3 scripts/validate_config_models.py; then
   exit 1
 fi
 
+# --- Preview the model-catalog seed ------------------------------------------
+# Deploys overwrite config/models from ui/definitions/models.json (the repo is
+# the source of truth), so a live console edit does not survive a deploy. Show
+# what this deploy's seed will change before the confirmation prompt below. On
+# a fresh project the Firestore API or database does not exist yet, so any
+# read failure only means "nothing to diff" -- the preview never blocks.
+echo "[>] Previewing the model-catalog seed (config/models)..."
+LIVE_MODELS_DOC=$(mktemp)
+if curl -sf \
+  "https://firestore.googleapis.com/v1/projects/${PROJECT}/databases/${FIRESTORE_DB_UI}/documents/config/models" \
+  -H "Authorization: Bearer $(gcloud auth application-default print-access-token 2>/dev/null || true)" \
+  -H "x-goog-user-project: ${PROJECT}" \
+  -o "$LIVE_MODELS_DOC" 2>/dev/null; then
+  python3 scripts/seed_config_models.py diff ui/definitions/models.json "$LIVE_MODELS_DOC"
+else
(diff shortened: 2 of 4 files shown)


deploy.sh
```diff
@@ -910,6 +930,23 @@ if [ "$CONFIG_SEED_STATUS" != "200" ]; then
...
 fi
 
+# The model catalog: config/models is overwritten from the repo file on every
+# deploy. Operators may edit the live doc between deploys; the pre-flight
+# preview above showed what this write replaces. Same fail-loudly contract as
+# the config/global seed.
+MODELS_SEED_STATUS=$(python3 scripts/seed_config_models.py convert < ui/definitions/models.json | curl -s -X PATCH \
+"https://firestore.googleapis.com/v1/projects/${PROJECT}/databases/${FIRESTORE_DB_UI}/documents/config/models" \
+  -H "Authorization: Bearer $(gcloud auth application-default print-access-token)" \
+  -H "x-goog-user-project: ${PROJECT}" \
```
victor-paunescu, line 940: Why is this needed, since we also pass the project in the URL?

DEVELOPING.md
```diff
@@ -58,6 +58,15 @@ curl -s -X PATCH \
   -H "x-goog-user-project: ${PROJECT}" \
   -H "Content-Type: application/json" \
   -d @<(envsubst < ./firestore_config_frontdoor.template.json)
+
+# The model catalog document (config/models), seeded from the repo file the
+# same way deploy.sh does it:
+python3 scripts/seed_config_models.py convert < ui/definitions/models.json | curl -s -X PATCH \
+  "https://firestore.googleapis.com/v1/projects/${PROJECT}/databases/${FIRESTORE_DB_UI}/documents/config/models" \
+  -H "Authorization: Bearer $(gcloud auth application-default print-access-token)" \
+  -H "x-goog-user-project: ${PROJECT}" \
```
victor-paunescu, line 67: same here

victor-paunescu (approved):

deploy.sh
christophervoelpel: ran this through a couple of hoops (incl. the PR I'm prepping for removing ADC):

The two projects answer different questions. The URL names the resource — which database this call writes. x-goog-user-project names the consumer — which project's quota the call is charged against and which project must have the Firestore API enabled.

With a bare user token over raw REST (which these curls use), the consumer does not default to the URL's project. A user token either carries its own quota project (wherever the operator's local credential state last pointed) or none at all, in which case the consumer falls back to the OAuth client's own project. So a call naming the right database in the URL can still 403 with "Cloud Firestore API has not been used in project X or it is disabled" for some X that isn't the target — or quietly charge another project's quota — depending on the machine it runs from.

The header pins the consumer to the deploy target so the deploy behaves identically regardless of local gcloud/ADC state. That's the standing contract in deploy.sh's header comment ("x-goog-user-project: $PROJECT on every curl, so it is runnable under CLOUDSDK_ACTIVE_CONFIG_NAME=<any config>"); this curl follows the same convention as the existing config/global and creative-template seeds above it.

DEVELOPING.md
christophervoelpel: same as above:

Same as in deploy.sh: the URL picks the database, the header picks the quota/API-enablement project. It matters most on a dev machine, where your personal credentials' quota project — if they have one at all — tends to point at whatever project you touched last, not the ${PROJECT} you just exported. The header makes the seed land its quota on the right project regardless.

PR #138, opened 2026-07-14
Keep output selection in sync with project

christophervoelpel:
## Summary

Keep output-video selection synchronized with asynchronously loaded project render runs.

## Behavior

- A direct route load selects the first available non-archived render after project data arrives.
- A valid explicit user selection survives updates to the same project.
- Removing or archiving the selected run chooses the first active fallback.
- Switching projects never carries a selection from the previous project.
- Existing playback and download behavior is unchanged.

## Tests

- UI compile and lint.
- Spec type-check.
- 23 test files / 207 tests, including delayed load, retained selection, removal and archive fallback, all-archived state, and project switching.
- Diff whitespace check.

## Risks / Notes

The change is confined to output selection state. It does not alter render persistence, backend APIs, playback, or downloads.

### DIFF
--- a/ui/src/app/output-video/output-video.ts
+++ b/ui/src/app/output-video/output-video.ts
@@ -22,6 +22,7 @@ import {
   computed,
   effect,
   inject,
+  linkedSignal,
   signal,
 } from '@angular/core';
 import {MatBadgeModule} from '@angular/material/badge';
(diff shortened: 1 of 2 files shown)


ui/src/app/output-video/output-video.ts
```diff
@@ -69,9 +70,30 @@ export class OutputVideo {
...
+    source: () => {
+      const config = this.configService.projectConfig.value();
+      return {
+        projectId: config.id,
+        renderRuns: config.renderRuns ?? [],
+      };
+    },
+    computation: (source, previous) => {
+      if (previous?.source.projectId === source.projectId && previous.value) {
+        const selectedTime = previous.value.createdAt.getTime();
```
victor-paunescu, line 86: I don't necessarily like that we use the creation time as an ID, but given that we don't have a better one, it should be fine.

victor-paunescu (approved):

PR #141, opened 2026-07-15
Retry runtime service account project IAM propagation

christophervoelpel:
## Summary

A fresh-project deploy can create and describe `sm-runtime` before project IAM accepts the account as a policy member. Retry the observed project-binding response at the failing write instead of treating service-account visibility as sufficient.

## Behavior

- Enables the longer propagation retry only for `serviceAccount:sm-runtime@<same-project>.iam.gserviceaccount.com` on a project IAM binding.
- Recognizes the expected missing-account response as a complete line even when `gcloud` adds non-error output around it.
- Fails immediately when the remaining output contains a permission, argument, missing-resource, quota, generic error, or conflict signal.
- Masks the expected account and project identifiers with literal replacements before legacy conflict matching, so names such as `stale-project` cannot cause retries.
- Uses 14 attempts for runtime-account propagation, with each backoff sleep capped at 60 seconds; ordinary IAM conflicts retain their six-attempt budget.

## Tests

- Covers recovery from the observed one-line and two-line responses, plus harmless warning and notice output before or after them.
- Covers immediate failure for mixed permission, quota, missing-resource, unrelated, and conflict responses, including after propagation has armed the longer budget.
- Covers permanent exhaustion, backoff bounds, identifier masking, unchanged legacy conflict behavior, and first-attempt success.
- Runs the backend suite across the supported Python versions, action signatures, formatting, Bash syntax, ShellCheck, and whitespace validation.

## Risks / Notes

- The expected missing-account sentence must still appear as an unchanged complete line. A same-line wording change stops deployment loudly and an idempotent rerun remains available.
- The longer budget deliberately applies only to the `sm-runtime` account created by this deploy. Supporting a configurable runtime-account name would require passing that created identity explicitly rather than inferring it from any IAM member.

### DIFF
--- a/deploy/libs.sh
+++ b/deploy/libs.sh
@@ -67,16 +67,30 @@ _retry_iam_write() {
       local policy_error_preamble
       local conflict_error
       local propagating_project
+      local propagation_residual
+      local line
       missing_sa_error="ERROR: (gcloud.projects.add-iam-policy-binding) INVALID_ARGUMENT: Service account ${propagating_runtime_sa} does not exist."
       policy_error_preamble='ERROR: Policy modification failed. For a binding with condition, run "gcloud alpha iam policies lint-condition" to identify issues in condition.'
       propagating_project="${propagating_runtime_sa#sm-runtime@}"
       propagating_project="${propagating_project%.iam.gserviceaccount.com}"
-      conflict_error="${err//${propagating_runtime_sa}/<runtime-sa>}"
-      conflict_error="${conflict_error//${propagating_project}/<project>}"
-      if [ "$err" = "$missing_sa_error" ] \
-          || [ "$err" = "${policy_error_preamble}"$'\n'"${missing_sa_error}" ]; then
-        retry_kind="runtime SA propagation"
-        propagation_seen=1
+      conflict_error="${err//"$propagating_runtime_sa"/<runtime-sa>}"
+      conflict_error="${conflict_error//"$propagating_project"/<project>}"
+      if grep -Fqx "$missing_sa_error" <<<"$err"; then
+        propagation_residual=""
+        while IFS= read -r line; do
+          if [ "$line" != "$missing_sa_error" ] \
+              && [ "$line" != "$policy_error_preamble" ]; then
+            propagation_residual="${propagation_residual}${line}"$'\n'
+          fi
+        done <<<"$err"
+        if ! grep -qiE \
+            'permission_denied|forbidden|unauthorized|invalid_argument|not_found|not found|does not exist|resource_exhausted|quota|concurrent|aborted|etag|stale' \
+            <<<"$propagation_residual" \
+            && ! grep -qE '^[[:space:]]*(ERROR|FATAL):' \
+              <<<"$propagation_residual"; then
+          retry_kind="runtime SA propagation"
+          propagation_seen=1
+        fi
       elif grep -qiE \
           'permission_denied|forbidden|unauthorized|invalid_argument|not_found|not found|does not exist' \
           <<<"$err"; then
--- a/deploy/libs.sh
+++ b/deploy/libs.sh
@@ -66,19 +66,24 @@ _retry_iam_write() {
       local missing_sa_error
       local policy_error_preamble
       local conflict_error
+      local propagating_project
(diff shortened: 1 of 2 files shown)


deploy/libs.sh
```diff
@@ -54,16 +61,47 @@ _retry_iam_write() {
...
+      local missing_sa_error
+      local policy_error_preamble
+      local conflict_error
+      local propagating_project
+      missing_sa_error="ERROR: (gcloud.projects.add-iam-policy-binding) INVALID_ARGUMENT: Service account ${propagating_runtime_sa} does not exist."
+      policy_error_preamble='ERROR: Policy modification failed. For a binding with condition, run "gcloud alpha iam policies lint-condition" to identify issues in condition.'
+      propagating_project="${propagating_runtime_sa#sm-runtime@}"
+      propagating_project="${propagating_project%.iam.gserviceaccount.com}"
+      conflict_error="${err//${propagating_runtime_sa}/<runtime-sa>}"
+      conflict_error="${conflict_error//${propagating_project}/<project>}"
```
victor-paunescu: In bash parameter expansion `${var//pattern/replace}`, unquoted pattern variables are evaluated as shell glob patterns (`*`, `?`, `[`). While GCP project IDs and SA emails use alphanumeric characters and hyphens, double-quoting pattern variables is recommended to guarantee literal string replacement:

```suggestion
      conflict_error="${err//"$propagating_runtime_sa"/<runtime-sa>}"
      conflict_error="${conflict_error//"$propagating_project"/<project>}"
```

deploy/libs.sh
```diff
@@ -54,16 +61,47 @@ _retry_iam_write() {
...
+      if [ "$err" = "$missing_sa_error" ] \
```
victor-paunescu: Exact equality comparison `[ "$err" = "$missing_sa_error" ]` is sensitive to minor formatting variations in `gcloud` output (e.g. CLI deprecation notices, extra preambles, or SDK updates). If `gcloud` adds any preamble lines, exact equality evaluates to `false` and falls through to line 80, matching `does not exist` or `INVALID_ARGUMENT` as a non-retryable error and failing immediately on attempt 1.

Consider using substring/regex matching for `Service account ${propagating_runtime_sa} does not exist.` instead of strict string equality.

deploy/libs.sh
```diff
@@ -137,6 +179,11 @@ add_iam_binding() {
   done
   local label="${role:+ (for $role)}"
   local project="$1"  # first positional arg
+  local propagating_runtime_sa=""
+  local expected_runtime_sa="sm-runtime@${project}.iam.gserviceaccount.com"
```
victor-paunescu, line 197: `expected_runtime_sa` is currently hardcoded to `sm-runtime@${project}.iam.gserviceaccount.com`. If a deploy uses a custom runtime service account name, `member` will not match `expected_runtime_sa`, leaving `propagating_runtime_sa=""` and bypassing the propagation retry budget.

Extracting the SA email dynamically from `--member=serviceAccount:...` when member matches `*.iam.gserviceaccount.com` would make propagation retries work transparently for custom service account names.

deploy/libs.sh
christophervoelpel: Thanks, agreed. Applied in 758cb88: both pattern variables are now quoted so the replacements are explicitly literal. Supported project IDs and service-account emails cannot contain Bash wildcard characters, so this does not change valid deployment behavior.

deploy/libs.sh
christophervoelpel: Agreed that exact whole-output equality lets harmless gcloud output disable the retry. Addressed in 758cb88 with the safety guard retained: the expected missing-account sentence must be a complete line, known expected lines are removed, and only remaining output without permission, resource, generic-error, or conflict signals is retryable. Added positive warning/notice cases and immediate-failure cases for PERMISSION_DENIED, RESOURCE_EXHAUSTED, and ABORTED output.

deploy/libs.sh
christophervoelpel: The deployment does not currently support a custom runtime service-account name: it always creates and uses sm-runtime. The narrow check intentionally gives the longer retry only to the account created moments earlier. Deriving it from any service-account member would also give deleted or mistyped accounts and Google-managed service agents the long retry budget. If custom names become supported, I would pass the created account explicitly rather than infer it from any member value, so I am keeping this scope unchanged.

victor-paunescu (approved):

PR #143, opened 2026-07-15
Isolate temporary files for concurrent video actions

christophervoelpel:
## Summary

Give each video conversion and combination call its own temporary directory, reuse identical inputs within a request, and upload completed outputs directly from disk. Concurrent requests cannot share local filenames or reuse stale probe results, while oversized input sets fail before download.

## Behavior

- Downloads each distinct Cloud Storage object once beneath a unique request directory and reuses that file for repeated arrangement entries.
- Rejects distinct inputs whose combined stored size exceeds 8 GiB before downloading any of them.
- Pins every download to the object generation measured during the size check, so replacing an object cannot bypass the aggregate limit.
- Uses a SHA-256 name plus an optional short alphanumeric extension for local inputs; each filename is at most 81 characters and does not copy object-path segments.
- Uploads converted and combined outputs from files while preserving their destination path, content type, and expiration metadata.
- Removes request-local inputs and outputs after both successful execution and errors.
- Removes the process-wide FFprobe path cache so each file is probed from its current contents.
- Does not change FFmpeg commands, codecs, MIME choices, or queue concurrency.

## Tests

- Covers simultaneous same-basename conversions and combinations, path-escape-like object names, thousand-character names and extensions, and request-local cleanup.
- Covers repeated large object references with one metadata read and one download, aggregate and single-input size rejection before FFmpeg, and generation-pinned downloads.
- Covers file-backed upload metadata and probing reused paths again.
- A current-head container smoke at `2e048469` ran two concurrent conversions and two reversed-order combinations through real Cloud Storage and FFmpeg 7.1.5. All outputs passed media, metadata, and cleanup checks; the temporary Job, objects, and image were removed afterward.

## Risks / Notes

- The 8 GiB limit applies per request. Output files and multiple concurrent requests can still exhaust the worker instance's in-memory filesystem.

### DIFF
--- a/actions/convert_video.py
+++ b/actions/convert_video.py
@@ -16,7 +16,6 @@
 
 from __future__ import annotations
 
-import pathlib
 import tempfile
 
 from actions_lib.ffmpeg import FFMPEG
@@ -25,6 +24,7 @@ from common import Key
 from common import NodeInput
 from common import NodeOutput
 from common import Params
+from util.gcs_wrapper import download_distinct_inputs
 from util.gcs_wrapper import GCS
 
 
@@ -47,14 +47,13 @@ def execute(
     A NodeOutput object containing the path to the converted video file.
   """
   file_path = file[0][Key.FILE.value]
-  basename = pathlib.PurePosixPath(file_path).name or 'input'
   with tempfile.TemporaryDirectory() as workdir:
-    local_path = pathlib.Path(workdir, f'0000_{basename}')
-    gcs.save_locally(file_path, str(local_path))
+    local_paths = download_distinct_inputs(gcs, [file_path], workdir)
+    local_path = local_paths[file_path]
 
     ffmpeg = FFMPEG().set_resolution(output_file_dimension)
     converted_video_path = ffmpeg.convert_video(
-        str(local_path), output_file_extension
+        local_path, output_file_extension
     )
 
     return {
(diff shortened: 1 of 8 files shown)


actions/convert_video.py
```diff
@@ -43,19 +47,20 @@ def execute(
...
     return {
-        "video": [{
-            Key.FILE.value: gcs.store(
-                output_file_bites,
-                f"output.{output_file_extension}",
+        'video': [{
+            Key.FILE.value: gcs.store_file(
+                converted_video_path,
+                f'output.{output_file_extension}',
                 ContentType.MP4.value,
```
victor-paunescu, line 64: Not sure how big of an issue this is: 
`ContentType.MP4.value` (`video/mp4`) is currently hardcoded here. If a user converts a video to a non-MP4 format (such as `webm`, `avi`, or `gif`), the uploaded GCS object will still be tagged with `video/mp4`.

christophervoelpel: Confirmed: this behavior already exists on main, and #143 intentionally leaves the conversion MIME contract unchanged while isolating temporary files. I am keeping that scope. The follow-up should validate a supported output extension and map it explicitly to the matching MIME type; unrestricted mimetypes.guess_type would accept containers the action has not proven it can produce.

victor-paunescu (approved):

PR #144, opened 2026-07-15
Recover transient post-action failures without rerunning completed work

christophervoelpel:
## Summary

Cloud Tasks may redeliver a worker task. Before this change, a transient
Firestore output write or successor-task creation failure after an action
completed could cause the action to run again, including another paid model
call. This change records task-specific completion so redelivery replays the
output and downstream work without rerunning the action; delivery count alone
never proves completion.

## Behavior

Two durable records use the workflow bucket and Firestore collection already
in the engine:

1. **Task completion output.** After an action completes, the worker creates
   `_task-completions/<task-token>.json` before writing the node output or
   creating successor tasks. The token hashes the Cloud Tasks queue and task
   names with the execution, node, and group, so it remains stable across
   redeliveries of that task. The object is create-only
   (`if_generation_match=0`) and read with one download. A redelivery that finds
   it replays the Firestore output write and successor-task creation without
   calling the action again. Completion objects carry no synthetic expiry
   metadata; retention follows the workflow bucket's lifecycle policy.

   The existing action cache cannot prove task completion: it is shared by
   action and inputs, and `forceExecution` deliberately bypasses it. The new
   object records the output of one specific task instead.

2. **Owner-fenced task state.** The existing Firestore lock document now records
   `running`, `retryable`, `succeeded`, or `failed`, together with the stable
   task token, a unique delivery-owner token, and a 1,860-second lease. A
   same-task redelivery while the owner is active receives a retryable response;
   an expired owner may be replaced in recovery-only mode; and a delayed owner
   cannot change its replacement's state. If the final success-state write
   fails, the handler makes a best-effort recovery-only transition before
   returning a retryable response; a success write that already committed
   cannot be reopened.

A missing completion object permits action execution only when durable state
explicitly records `recoveryOnly: false`. This covers a new task, a quota 429,
and a transient completion download that occurred before any object bytes were
observed. Post-action and ambiguous completion failures remain recovery-only:
later deliveries may replay saved output but may not call the action, and a
missing object or transient download failure returns a retryable response.

Once completion bytes are downloaded, malformed JSON, invalid encoding,
excessive nesting, or a non-object root fails terminally on the first observed
delivery. Unexpected local parser failures remain retryable but recovery-only,
so a later missing object cannot permit the action to run.

If the action cache or completion-object write fails after output was produced,
the task stops rather than risking another paid call. The handler uses the
deployed queues' 30-attempt limit. Quota 429 errors, including
client-library-wrapped forms, remain retryable; model 5xx errors and timeouts are
not retried automatically.

`ROLE=worker` requires the Cloud Tasks queue name, task name, and retry-count
headers before lock access. These headers identify the delivery; Cloud Run
IAM/OIDC remains the authentication boundary. `ROLE=all` keeps the headerless
local-development path.

## Tests

Covers complete, missing, empty, partial, malformed, and local delivery headers;
nonzero retry counts on a task's first observed delivery; forced-action quota
recovery; transient initial completion reads with absent and present output;
create-only write conflicts and single-request reads; terminal invalid encoding,
JSON, nesting, and root shape; recovery-only unexpected parser failures and
create-conflict read failures; missing recovery output; absence of synthetic
expiry metadata; malformed recovery state; active and expired leases; owner
fencing; definite and ambiguous state-write failures; action-cache and
completion-write failures; wrapped quota errors; the 30-attempt boundary; and
ambiguous successor-task creation.

## Risks / Notes

- A model call and its completion-object write cannot be atomic. If output was
  produced but durable completion cannot be proven, the task stops rather than
  risk another billable action.
- If both the success-state write and its best-effort retryable transition fail,
  the document remains `running` until its lease expires. The configured queue
  retry schedule outlives that lease.
- If a fatal action error occurs and Firestore also cannot record `failed`, the
  handler still acknowledges the task to avoid a potentially billable retry;
  the task document may remain `running` until Firestore TTL cleanup.
- For this protocol, a claimed delivery performs one completion-object read. A
  delivery that executes an action also attempts one create-only object write,
  and a claimed delivery normally performs one fenced state-transition
  transaction after the existing claim transaction.
- Create-only writes prevent this worker protocol from replacing an existing
  completion object. A principal with bucket write permission can still modify
  or delete it.
- Completion objects remain until the workflow bucket's lifecycle policy
  removes them; this repository does not install that policy.
- An ambiguous successor-task response may create a duplicate delivery; the
  existing join seal prevents that duplicate from releasing the downstream node
  twice.
- This change does not add a general workflow watchdog or UI repair path.

### DIFF
--- a/orch.py
+++ b/orch.py
@@ -324,12 +324,44 @@ def trigger_action_handler() -> tuple[str, int]:
         node,
     )
     return 'Already Triggered', 200
+  task_data = copy.deepcopy(data)
   try:
     orchestrator.trigger_action(
-        copy.deepcopy(data),
+        task_data,
         instance,
         retry_count < _MAX_ALLOWED_RETRIES,
+        retry_count > 0,
     )
+  except util_errors.RetryableTaskRecoveryError as e:
+    if retry_count < _MAX_ALLOWED_RETRIES:
+      logger.error(
+          'Retrying forced action cache recovery for %s: %s',
+          data[Key.ACTION.value],
+          e.__cause__ or e,
+      )
+      orchestrator.db.release_task_lock(execution_id, node_id, group_id)
+      return 'Service Unavailable', 503
+    logger.error(
+        'Retry limit reached during forced action cache recovery for %s: %s',
+        data[Key.ACTION.value],
+        e.__cause__ or e,
+    )
+    return 'Internal Error', 200
+  except util_errors.RetryablePostActionError as e:
+    if retry_count < _MAX_ALLOWED_RETRIES:
+      logger.error(
+          'Retrying completed action %s after infrastructure failure: %s',
+          data[Key.ACTION.value],
+          e.__cause__ or e,
+      )
+      orchestrator.db.release_task_lock(execution_id, node_id, group_id)
+      return 'Service Unavailable', 503
+    logger.error(
+        'Retry limit reached after completed action %s: %s',
+        data[Key.ACTION.value],
+        e.__cause__ or e,
+    )
+    return 'Internal Error', 200
   except Exception as e:  # pylint: disable=broad-exception-caught
     if util_errors.is_retryable(e):
       logger.error('Retrying action %s: %s', data[Key.ACTION.value], e)
--- a/orch.py
+++ b/orch.py
@@ -308,10 +308,18 @@ def trigger_action_handler() -> tuple[str, int]:
     # Malformed task payload (e.g. missing the Cloud Tasks lock fields): reject
     # cleanly instead of raising a KeyError below and returning a 500.
     return 'Bad Request', 400
+  queue_name = flask_request.headers.get('X-CloudTasks-QueueName')
+  task_name = flask_request.headers.get('X-CloudTasks-TaskName')
+  retry_count_header = flask_request.headers.get('X-CloudTasks-TaskRetryCount')
+  task_headers = (queue_name, task_name, retry_count_header)
+  task_header_presence = tuple(value is not None for value in task_headers)
+  if _ROLE == 'worker':
+    if not all(task_headers):
+      return 'Bad Request', 400
+  elif any(task_header_presence) and not all(task_headers):
+    return 'Bad Request', 400
   try:
-    retry_count = int(
-        flask_request.headers.get('X-CloudTasks-TaskRetryCount', 0)
-    )
+    retry_count = int(retry_count_header or 0)
   except (TypeError, ValueError):
     return 'Bad Request', 400
   if retry_count < 0:
@@ -324,10 +332,6 @@ def trigger_action_handler() -> tuple[str, int]:
   node_id = data[Key.NODE_ID.value]
   group_id = data[Key.GROUP_ID.value]
   node = data[Key.WORKFLOW_DEF.value][node_id]
-  queue_name = flask_request.headers.get('X-CloudTasks-QueueName')
-  task_name = flask_request.headers.get('X-CloudTasks-TaskName')
-  if bool(queue_name) != bool(task_name):
-    return 'Bad Request', 400
   task_identity = (
       ['cloud-tasks', queue_name, task_name]
       if queue_name and task_name
@@ -487,6 +491,7 @@ def trigger_action_handler() -> tuple[str, int]:
         data[Key.ACTION.value],
         e,
     )
+    mark_retryable(e, True)
     return 'Service Unavailable', 503
   if not task_succeeded:
     logger.error(
--- a/orch.py
+++ b/orch.py
@@ -43,6 +43,7 @@ optional; the defaults reproduce the single-service behavior exactly):
 
 import copy
 import datetime
+import hashlib
 import json
 import logging
 import os
@@ -70,8 +71,8 @@ from util import errors as util_errors
 from util import submission_validation
 from werkzeug.security import safe_join
 
-# At most the queue's allowed attempts minus one, so the workflow proceeds:
-_MAX_ALLOWED_RETRIES = 10
+# deploy.sh configures every worker queue with 30 total attempts.
+_MAX_TASK_ATTEMPTS = 30
 
 _ROLE = os.environ.get('ROLE', 'all')
 _AUTH_MODE = os.environ.get('AUTH_MODE', 'none')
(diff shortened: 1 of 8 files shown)


orch.py
```diff
@@ -324,21 +373,127 @@ def trigger_action_handler() -> tuple[str, int]:
...
+      )
+      if mark_failed(e):
+        return 'Internal Error', 200
+      return 'Service Unavailable', 503
     else:
       logger.error('Fatal error for action %s: %s', data[Key.ACTION.value], e)
+      mark_failed(e)
       return 'Internal Error', 200  # Cloud Tasks will NOT retry this
+  try:
+    task_succeeded = orchestrator.db.mark_task_succeeded(
```
victor-paunescu, line 480: When `orchestrator.db.mark_task_succeeded(...)` fails due to a transient Firestore network error, `orch.py` catches the exception (line 484) and returns HTTP `503 Service Unavailable`.
However, `mark_retryable` is **not** called in this exception block. The Firestore task document remains in `_TASK_STATE_RUNNING` with `leaseExpiresAt = now + 1860s` (31 minutes). 
When Cloud Tasks receives the 503 response and retries seconds later, `acquire_task_lock` checks:
```python
if current_state != _TASK_STATE_RETRYABLE and not lease_expired:
  return TASK_LOCK_BUSY
```
Because only a few seconds have passed, `lease_expired` is `False`, so `acquire_task_lock` returns `TASK_LOCK_BUSY`. `orch.py` responds with 503 for **every subsequent retry attempt** until Cloud Tasks exhausts its 30 max attempts (~10-15 minutes), dropping the task completely before the 31-minute lease ever expires!
**Suggested Fix:**
In the exception block for `mark_task_succeeded`, explicitly call `mark_retryable(e, recovery_only=True)` before returning 503

orch.py
```diff
@@ -307,14 +308,62 @@ def trigger_action_handler() -> tuple[str, int]:
...
   execution_id = data[Key.EXECUTION_ID.value]
   node_id = data[Key.NODE_ID.value]
   group_id = data[Key.GROUP_ID.value]
   node = data[Key.WORKFLOW_DEF.value][node_id]
-  if not orchestrator.db.acquire_task_lock(execution_id, node_id, group_id):
+  queue_name = flask_request.headers.get('X-CloudTasks-QueueName')
+  task_name = flask_request.headers.get('X-CloudTasks-TaskName')
+  if bool(queue_name) != bool(task_name):
+    return 'Bad Request', 400
+  task_identity = (
```
victor-paunescu, line 335: `task_identity` degrades to `['local', execution_id, node_id, group_id]` if `X-CloudTasks-QueueName` or `X-CloudTasks-TaskName` headers are omitted.
Because Firestore task lock documents are indexed by `(execution_id, node_id, group_id)` regardless of `task_token` (`util/database.py#L393`), an unauthenticated or test HTTP request without Cloud Tasks headers will acquire the lock using `taskToken = sha256(['local', ...])`.
When the legitimate Cloud Task arrives with headers set, `acquire_task_lock` compares `current_task_token != task_token` (`cloud-tasks` vs `local`), detects a token mismatch, and returns `TASK_LOCK_DUPLICATE`. `orch.py` translates this to `'Already Triggered', 200 OK`, silently dropping execution of the real Cloud Task.
**Suggested Fix:** Enforce mandatory Cloud Tasks headers when running in non-local production worker mode

orchestrator.py
```diff
@@ -268,25 +274,127 @@ def _inform_successors(
...
+  return gcs.gcs_bucket.blob(f'{_TASK_COMPLETION_PREFIX}/{task_token}.json')
+
+
+def _load_task_completion(
+    data: dict[str, Any], task_token: str, recovery_only: bool = False
+) -> dict[str, Any] | None:
+  """Loads an immutable task output without falling back to execution."""
+  blob = _task_completion_blob(data, task_token)
+  try:
+    blob.reload()
```
victor-paunescu: `_load_task_completion` executes `blob.reload()` followed by `blob.download_as_string(if_generation_match=generation)`, incurring **two separate GCS network round-trips** on every completion probe.
`download_as_string()` already fetches object content from GCS directly. If the blob does not exist, `download_as_string()` raises `google_exceptions.NotFound`.
**Suggested Fix:** Remove `blob.reload()` and call `download_as_string()` directly, capturing the generation from the downloaded metadata or catching `NotFound`

victor-paunescu (changes requested):

orchestrator.py
```diff
@@ -268,25 +274,127 @@ def _inform_successors(
...
+    ) from e
+
+
+def _store_task_completion(
+    data: dict[str, Any], task_token: str, output: dict[str, Any]
+) -> dict[str, Any]:
+  """Creates the immutable output used by later task deliveries."""
+  blob = _task_completion_blob(data, task_token)
+  blob.metadata = {
+      'timeToDelete': (
```
victor-paunescu: Google Cloud Storage Native Object Lifecycle Management does not automatically delete objects based on a custom dictionary key named `timeToDelete`. While Firestore natively supports TTL fields like this, GCS requires either `DaysSinceCustomTime` (using blob.custom_time) or an out-of-band Cloud Function/sweeper to process it.

orch.py
christophervoelpel: Implemented in 4c2e9d4. A failed success-state write now makes a best-effort, owner-fenced transition to retryable with recoveryOnly=true before returning 503. The next delivery replays the manifest without running the action. Tests cover both a definite write failure and a commit-then-error response; a committed succeeded state cannot be reopened. One timing correction: the queue delay increases linearly after three doublings, so the configured retry budget already outlives the lease. Output and successors had already completed on this path; this change enables immediate recovery and prevents stale running state.

orchestrator.py
christophervoelpel: Implemented in 4c2e9d4. Completion probes now use one download_as_bytes() request and handle NotFound directly. The separate metadata read and unanchored generation check were removed; create-only writes still use if_generation_match=0. Tests pin one content request per probe.

orchestrator.py
christophervoelpel: Confirmed and fixed in 4c2e9d4. Completion manifests no longer carry timeToDelete, and the unused manifest TTL constant was removed. They now follow the bucket lifecycle policy described in README. This change leaves existing action-output metadata and operator-managed bucket rules untouched.

orch.py
christophervoelpel: Implemented in 4c2e9d4. ROLE=worker now requires queue name, task name, and retry count before any task lock access; missing, empty, partial, invalid, or negative header sets return 400. ROLE=all retains the headerless local path. These headers identify the delivery; Cloud Run IAM/OIDC remains the authentication boundary.

orchestrator.py
```diff
@@ -268,25 +273,108 @@ def _inform_successors(
...
+  except Exception as e:  # pylint: disable=broad-exception-caught
+    raise util_errors.RetryableTaskRecoveryError(
+        'Failed to read task completion output'
+    ) from e
+  try:
+    output = json.loads(manifest)
+    if not isinstance(output, dict):
+      raise TypeError('Task completion output must be a JSON object')
+    return output
+  except Exception as e:  # pylint: disable=broad-exception-caught
```
victor-paunescu, line 346: Catching broad `Exception` here wraps `json.JSONDecodeError` or `TypeError` into a `RetryableTaskRecoveryError`, causing Cloud Tasks to retry 30 times on corrupted data. Consider catching `(json.JSONDecodeError, TypeError)` specifically to fail terminally when data is unparseable.

victor-paunescu (changes requested):

orchestrator.py
christophervoelpel: Confirmed and fixed in 65d543a. Successfully downloaded completion content that cannot be decoded or validated now fails terminally on its first observed delivery. The terminal cases cover malformed JSON, invalid encoding, excessive nesting, and a non-object root. GCS download failures and recovery-only NotFound remain retryable; unexpected local parser failures retry recovery-only, so disappearance of an already-observed object cannot permit the action to run. The regressions assert first-attempt terminal state, no action or output write, and one completion read.

victor-paunescu (approved):

PR #145, opened 2026-07-16
Serve the model catalog to UI dropdowns

christophervoelpel:
## Summary

Serve the model catalog through the existing `/api/config` response and use it for video-model choices on the setup and storyboard pages. This removes the hardcoded UI list so displayed choices follow the same catalog the backend validates.

## Behavior

- `/api/config` adds `modelCatalog` and `modelCatalogSource` while preserving the existing global configuration fields.
- Video-model dropdowns list models that support `generate_video` at the configured Veo location. Model IDs are sorted for stable ordering across live and shipped catalog sources.
- When the live catalog no longer contains a saved project model, the configuration service switches it to the compatible Veo default and shows a notice on any project page. If that default is unavailable at the configured location, it uses the first compatible model.
- A new project whose deploy default is no longer available uses the compatible catalog model before its first save. Saved-project corrections remain visible and are persisted.
- Missing catalogs and the shipped fallback never rewrite existing projects. An unusable live catalog keeps `/api/config` available with the shipped catalog.
- The hardcoded UI model constant and its duplicate drift check are removed.

## Tests

- Backend coverage verifies live-catalog responses, shipped fallback behavior, preservation of global configuration fields, and a console-edited catalog containing Firestore timestamp values.
- UI coverage verifies action and location filtering, stable ordering, missing-config behavior, saved and unsaved model corrections, the immediate new-project payload, incompatible defaults, visible notices, and no rewrite from the shipped fallback.
- Setup and storyboard component fixtures use the catalog-derived model list, while the remaining shipped-model check continues to verify the deployment template against the allowlist.

## Risks / Notes

- Dropdowns are empty while `/api/config` is loading or unavailable.
- Catalog changes reach an open browser tab on reload; this keeps the existing freshness behavior of global configuration.
- Removing a model from the live catalog can update saved projects that used it. The UI announces that update and does not perform it from the shipped fallback.

### DIFF
--- a/ui/src/app/services/config/config.ts
+++ b/ui/src/app/services/config/config.ts
@@ -58,13 +58,21 @@ export interface GcsFile {
 }
 
 /**
- * List of available video generation models.
+ * One model's entry in the catalog served by /api/config (the same catalog
+ * the backend validator enforces, so the dropdowns cannot drift from it).
  */
-export const VIDEO_GENERATION_MODELS = [
-  'veo-3.1-generate-001',
-  'veo-3.1-fast-generate-001',
-  'veo-3.1-lite-generate-001',
-];
+export interface ModelCatalogEntry {
+  family: string;
+  actions: string[];
+  locations: string[];
+  capabilities?: Record<string, unknown>;
+}
+
+export interface ModelCatalog {
+  defaults: Record<string, string>;
+  actions: Record<string, {location_param: string | null; default_key: string}>;
+  models: Record<string, ModelCatalogEntry>;
+}
 
 interface GlobalConfig {
   // GCP
@@ -95,6 +103,11 @@ interface GlobalConfig {
   // FFmpeg
   encodingSpeed: number;
   qualityLevel: number;
+
+  // Model catalog (merged into the response by the backend; 'firestore' when
+  // the live config/models doc was served, 'shipped' on fallback).
+  modelCatalog?: ModelCatalog;
+  modelCatalogSource?: 'firestore' | 'shipped';
 }
 
 /**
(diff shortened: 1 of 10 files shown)


evil-shrike (approved):

ui/src/app/services/config/config.ts
```diff
@@ -507,6 +559,39 @@ export class ConfigService {
       );
       this.document.documentElement.classList.add(this.primaryColor());
     });
+    // Swap a project's model to the catalog default when the catalog no
+    // longer offers it. Service-level: reopened projects land on the
```
victor-paunescu: not sure about this comment
`Service-level: reopened projects land on the storyboard page, not setup.`

Reopened projects land wherever you go to, right? If we open /{id}/setup, it will land on setup.

ui/src/app/services/config/config.ts
```diff
@@ -507,6 +559,39 @@ export class ConfigService {
...
+      const fallback = this.liveCatalogVideoFallback();
+      if (!project.id || !fallback) {
+        return;
+      }
+      if (project.model && models.includes(project.model)) {
+        return;
+      }
+      const previous = project.model;
+      // Persistence decides the branch: a saved project sitting on the
+      // deploy default must still get the visible, persisted rewrite.
```
victor-paunescu: What branch?

ui/src/app/services/config/config.ts
```diff
@@ -696,12 +781,17 @@ export class ConfigService {
...
     this.persistedProjectIds.delete(uuid);
-    this.projectConfig.set({
+    const project = {
       ...this.DEFAULT_PROJECT_CONFIG(),
       id: uuid,
       name: 'Untitled Project',
       createdBy: undefined,
-    });
+    };
+    const fallback = this.liveCatalogVideoFallback();
```
victor-paunescu: Why get the fallback before checking `!this.videoModels().includes(project.model)`?

ui/src/app/services/config/config.ts
christophervoelpel: Correct. Direct /:id/setup opens Setup; the homepage card only happens to open storyboard. I replaced the overbroad route claim with the actual invariant: unavailable-model correction stays in the shared service because not every project route creates the setup component. No behavior changed here.

ui/src/app/services/config/config.ts
christophervoelpel: Branch meant the two update paths below, but that wording was opaque. I removed it and made the names explicit: a server-confirmed project gets a visible persisted correction; a never-saved project on the deploy default is corrected without creating a project solely due to the catalog change. I also covered the reachable empty-model case so a malformed saved project is PATCHed instead of being repaired only in memory.

ui/src/app/services/config/config.ts
christophervoelpel: No behavioral reason in setNewProject. It is a plain method, so I now check membership first and derive a fallback only when the configured model is unavailable. The constructor effect keeps its existing up-front reactive reads. Tests now cover both directions: an unavailable configured default is replaced before POST, and a compatible configured default is retained.

victor-paunescu (approved):

PR #147, opened 2026-07-17
Serialize project saves and enforce render eligibility

christophervoelpel:
## Summary

Serialize whole-project saves per project and use one frame-safe render
contract from storyboard editing through arrangement submission.

## Behavior

- Only one save request per project can be in flight. Newer edits wait behind
  it, and intermediate queued snapshots are replaced by the latest state.
- Queued payloads are copied after their edit timestamp is set, so later nested
  mutations cannot change already queued bytes.
- A reload begun during an unsettled save preserves the newest local snapshot,
  including stale GET and 404 responses. Save state remains isolated by project
  and is removed after project deletion.
- Retry sends the newest observed project snapshot and does nothing when that
  snapshot has already saved.
- Save state is dropped when a delete begins rather than when it returns, so a
  save completing mid-deletion cannot dispatch its queued successor. A request
  already sent may still reach the server after the delete; that ordering is
  the server's.
- Filmstrip, preview, and render submission use only clips with a non-empty
  Cloud Storage path, at least 0.042 seconds of effective duration, and a trim
  that lies inside the source video.
- Trim edits wait for video metadata and keep millisecond resolution: endpoints
  convert to integer milliseconds before any minimum-duration or ordering
  repair.
- An invalid intended clip disables Render with a recoverable message rather
  than silently producing a partial storyboard. A render already in progress
  reports that instead.
- A transition longer than either clip it joins disables Render and is refused
  at arrangement construction, since the crossfade would consume a neighbouring
  clip entirely.
- An explicit zero transition overlap means a hard cut end to end; only an
  absent overlap takes the default.
- Arrangement construction uses the same resolved clip values and the same
  transition check as eligibility and preview.

## Tests

- Covers delayed saves, three-edit coalescing, payload isolation, navigation
  during a save, stale and missing reload responses, create-conflict recovery,
  retry, cleanup, and catalog correction.
- Covers crossed and sub-frame trims, millisecond precision, trims starting at
  or running past the source end, edits before metadata loads, URL-only and
  missing-duration media, invalid mixed storyboards, oversized transitions in
  both the UI and arrangement construction, explicit zero overlap through to
  the backend, and exact arrangement values.
- The exact combined head passes the full UI build, lint, unit-test, spec
  type-check, and repository checks.

## Risks / Notes

- Concurrent browser tabs and multiple users remain last-writer-wins; save
  serialization is scoped to one app session. Ordering a save against a delete
  is likewise client-side only: a request already in flight when the delete is
  issued can still be applied server-side afterwards.
- Legacy URL-only media must be re-uploaded or repaired before rendering.

### DIFF
--- a/ui/src/app/composition/composition.ts
+++ b/ui/src/app/composition/composition.ts
@@ -148,6 +148,9 @@ export class Composition {
   );
 
   renderDisabledReason = computed(() => {
+    if (this.combiningScenes()) {
+      return 'Video rendering is in progress.';
+    }
     if (
       this.sceneRenderClips().some(
         ({resolution}) => resolution.state === 'invalid',
--- a/ui/src/app/composition/composition.ts
+++ b/ui/src/app/composition/composition.ts
@@ -39,6 +39,7 @@ import {FormatTimePipe} from '../pipes/format-time-pipe';
 import {
   ConfigService,
   DEFAULT_TRANSITION_OVERLAP,
+  findTransitionContractViolation,
   resolveSceneRenderClip,
 } from '../services/config/config';
 import {MediaRef, MediaService} from '../services/media/media';
@@ -139,12 +140,17 @@ export class Composition {
    * clip is valid. An empty playlist would fail in the backend, while omitting
    * an invalid intended clip would silently render only part of the storyboard.
    */
+  private transitionViolation = computed(() =>
+    findTransitionContractViolation(this.scenes()),
+  );
+
   canRender = computed(
     () =>
       this.playlist().length > 0 &&
       !this.sceneRenderClips().some(
         ({resolution}) => resolution.state === 'invalid',
-      ),
+      ) &&
+      this.transitionViolation() === null,
   );
 
   renderDisabledReason = computed(() => {
@@ -163,6 +169,10 @@ export class Composition {
     }
     if (this.playlist().length === 0) {
       return 'Select or upload at least one scene video before rendering.';
+    }
+    const transitionViolation = this.transitionViolation();
+    if (transitionViolation) {
+      return transitionViolation;
     }
     return '';
   });
--- a/ui/src/app/composition/composition.ts
+++ b/ui/src/app/composition/composition.ts
@@ -39,6 +39,7 @@ import {FormatTimePipe} from '../pipes/format-time-pipe';
 import {
   ConfigService,
   DEFAULT_TRANSITION_OVERLAP,
+  resolveSceneRenderClip,
 } from '../services/config/config';
 import {MediaRef, MediaService} from '../services/media/media';
 import {MediaSrcPipe} from '../services/media/media-src.pipe';
@@ -99,74 +100,69 @@ export class Composition {
     return ratio ? ratio.replace(':', '/') : '16/9';
   });
 
-  filmstripScenes = computed(() => {
-    return this.scenes().filter(scene => {
-      if (
-        this.configService.isGeneratedScene(scene) &&
-        scene.candidates &&
-        scene.selectedCandidateIndex !== undefined
-      ) {
-        // Downstream resolves media by path (signing it on read), so a scene
-        // persisted with a path but a blank url is still renderable: accept
-        // either a path or a url.
-        const video = scene.candidates[scene.selectedCandidateIndex].video;
-        return !!(video?.path || video?.url);
+  private sceneRenderClips = computed(() =>
+    this.scenes().map(scene => ({
+      scene,
+      resolution: resolveSceneRenderClip(scene),
+    })),
+  );
+
+  filmstripScenes = computed(() =>
+    this.sceneRenderClips()
+      .filter(({resolution}) => resolution.state === 'ready')
+      .map(({scene}) => scene),
+  );
+
+  playlist = computed(() => {
+    return this.sceneRenderClips().flatMap(({scene, resolution}) => {
+      if (resolution.state !== 'ready') {
+        return [];
       }
-      if (this.configService.isProvidedVideoScene(scene)) {
-        return !!(scene.video?.path || scene.video?.url);
-      }
-      return false;
+      const {video, start, duration} = resolution.clip;
+      return [
+        {
+          id: scene.id,
+          name: scene.name,
+          video,
+          start,
+          end: start + duration,
+          duration,
+          transitionOverlap: scene.transitionOverlap,
+          type: scene.type,
+        },
+      ];
     });
   });
 
-  playlist = computed(() => {
-    return this.filmstripScenes()
-      .map(scene => {
-        if (this.configService.isGeneratedScene(scene)) {
-          if (!scene.candidates || scene.selectedCandidateIndex === undefined) {
-            return;
-          }
-          const candidate = scene.candidates[scene.selectedCandidateIndex];
-          const start = candidate.trim?.start ?? 0;
-          const end = candidate.trim?.end ?? candidate.durationSeconds;
-          return {
-            id: scene.id,
-            name: scene.name,
-            video: candidate.video,
-            start,
-            end,
-            duration: end - start,
-            transitionOverlap: scene.transitionOverlap,
-            type: 'generated' as const,
-          };
-        } else {
-          // VideoScene
-          const start = scene.trim?.start ?? 0;
-          const end = scene.trim?.end ?? scene.durationSeconds!;
-          return {
-            id: scene.id,
-            name: scene.name,
-            video: scene.video,
-            start,
-            end,
-            duration: end - start,
-            transitionOverlap: scene.transitionOverlap,
-            type: 'video' as const,
-          };
-        }
-      })
-      .filter(item => item !== undefined);
-  });
-
   /**
-   * True when at least one scene contributes a usable video to the render
-   * (a selected candidate with a video, or an uploaded video scene). The
-   * playlist already filters to exactly those scenes, so an empty playlist
-   * means there is nothing to combine — rendering would fail in the backend
-   * with "cannot generate video without at least one video input", so we block
-   * it in the UI instead.
+   * True when at least one scene contributes a usable video and every intended
+   * clip is valid. An empty playlist would fail in the backend, while omitting
+   * an invalid intended clip would silently render only part of the storyboard.
    */
-  canRender = computed(() => this.playlist().length > 0);
+  canRender = computed(
+    () =>
+      this.playlist().length > 0 &&
+      !this.sceneRenderClips().some(
+        ({resolution}) => resolution.state === 'invalid',
+      ),
+  );
+
+  renderDisabledReason = computed(() => {
+    if (
+      this.sceneRenderClips().some(
+        ({resolution}) => resolution.state === 'invalid',
+      )
+    ) {
+      return (
+        'One or more selected scene videos are missing a storage path or valid ' +
+        'duration, or have an invalid trim.'
+      );
+    }
+    if (this.playlist().length === 0) {
+      return 'Select or upload at least one scene video before rendering.';
+    }
+    return '';
+  });
 
   currentPlaylistIndex = signal(0);
   isPlaying = signal(false);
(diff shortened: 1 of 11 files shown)


ui/src/app/services/config/config.ts
```diff
@@ -653,44 +799,50 @@ export class ConfigService {
...
+    const isPersisted = this.persistedProjectIds.has(projectId);
     const request = isPersisted
-      ? this.httpClient.patch(`/api/projects/${config.id}`, config)
-      : this.httpClient.post<{id: string}>('/api/projects', config);
+      ? this.httpClient.patch(`/api/projects/${projectId}`, save.payload)
+      : this.httpClient.post<{id: string}>('/api/projects', save.payload);
     request.subscribe({
       next: () => {
-        this.persistedProjectIds.add(config.id);
+        this.persistedProjectIds.add(projectId);
```
victor-paunescu, line 895: If `deleteProject(projectId)` is called while a save request is in flight, `deleteProject` removes `projectId` from `persistedProjectIds` and deletes its `ProjectSaveState`. When the in-flight HTTP request succeeds, `this.persistedProjectIds.add(projectId)` unconditionally re-inserts the deleted project ID into `persistedProjectIds`. Guarding this with `this.projectSaveStates.has(projectId)` prevents resurrecting deleted project IDs.

```suggestion
      next: () => {
        if (!this.projectSaveStates.has(projectId)) {
          return;
        }
        this.persistedProjectIds.add(projectId);
```

ui/src/app/services/config/config.ts
```diff
@@ -700,13 +852,16 @@ export class ConfigService {
...
+          // Retry the current state when still on this project, otherwise the
+          // newest source this project's queue has observed.
           const latest = this.projectConfig.value();
-          this.persistNow(latest.id === config.id ? latest : config);
+          this.persistNow(
+            latest.id === projectId
+              ? latest
+              : (state.latestSource ?? save.source),
+          );
         });
```
victor-paunescu, line 954: If a save fails and the user clicks "Retry" on the error snackbar after deleting the project, `persistNow` calls `this.saveState(config.id)`, which instantiates a new `ProjectSaveState` for a deleted project. Checking `this.projectSaveStates.has(projectId)` inside `onAction()` avoids recreating state for deleted projects.

```suggestion
        snackBarRef.onAction().subscribe(() => {
          if (!this.projectSaveStates.has(projectId)) {
            return;
          }
          // Retry the current state when still on this project, otherwise the
          // newest source this project's queue has observed.
          const latest = this.projectConfig.value();
          this.persistNow(
            latest.id === projectId
              ? latest
              : (state.latestSource ?? save.source),
          );
        });
```

ui/src/app/storyboard/storyboard.ts
```diff
@@ -599,6 +603,43 @@ export class Storyboard {
...
     newTrim.end = toDecimals(newTrim.end, 3);
 
+    const minimumDurationMilliseconds = Math.round(
+      MIN_RENDER_CLIP_DURATION_SECONDS * 1000,
+    );
+    const sourceDurationMilliseconds = Math.round(
+      toDecimals(duration, 3) * 1000,
+    );
+    let startMilliseconds = Math.round(newTrim.start * 1000);
+    let endMilliseconds = Math.round(newTrim.end * 1000);
```
victor-paunescu: Calling `toDecimals` prior to `Math.round(val * 1000)` introduces potential IEEE 754 precision loss (`1.001 * 1000` evaluating to `1000.9999999999999` which `Math.floor` inside `toDecimals` truncates to `1000`, losing 1ms). Direct integer conversion via `Math.round` avoids truncation, and lines 640-641 convert back to exact 3-decimal seconds cleanly.

```suggestion
    const sourceDurationMilliseconds = Math.round(duration * 1000);
    let startMilliseconds = Math.round(newTrim.start * 1000);
    let endMilliseconds = Math.round(newTrim.end * 1000);
```

ui/src/app/composition/composition.ts
```diff
@@ -99,74 +100,69 @@ export class Composition {
...
+      return (
+        'One or more selected scene videos are missing a storage path or valid ' +
+        'duration, or have an invalid trim.'
+      );
+    }
+    if (this.playlist().length === 0) {
+      return 'Select or upload at least one scene video before rendering.';
+    }
+    return '';
+  });
```
victor-paunescu, line 178: In `composition.html`, the Render button is disabled when `combiningScenes()` is true. Adding a check for `this.combiningScenes()` in `renderDisabledReason` ensures users see an informative tooltip explanation on hover while rendering is active.

```suggestion
  renderDisabledReason = computed(() => {
    if (this.combiningScenes()) {
      return 'Video rendering is in progress.';
    }
    if (
      this.sceneRenderClips().some(
        ({resolution}) => resolution.state === 'invalid',
      )
    ) {
      return (
        'One or more selected scene videos are missing a storage path or valid ' +
        'duration, or have an invalid trim.'
      );
    }
    if (this.playlist().length === 0) {
      return 'Select or upload at least one scene video before rendering.';
    }
    return '';
  });
```

ui/src/app/services/remix-engine/remix-engine-mediated.spec.ts
```diff
@@ -1092,18 +1101,122 @@ describe('RemixEngineService (mediated)', () => {
     });
   });
 
+  describe('combineScenes: render contract', () => {
```
victor-paunescu, line 1104: Adding an explicit assertion for storyboards containing only unselected generated scenes (`resolution.state === 'not-selected'`) verifies that `combineScenes` correctly triggers the contract error snackbar when no ready scenes exist.
```suggestion
    it('does not submit a workflow when all generated scenes are unselected', async () => {
      setRenderStoryboard([
        {id: 'gen-1', type: 'generated', name: 'Unselected Scene'},
      ]);
      const startSpy = vi.spyOn(service, 'startCombineScenesWorkflow');
      await service.combineScenes();
      expect(startSpy).not.toHaveBeenCalled();
      expect(matSnackBarMock.open).toHaveBeenCalledWith(
        'Select or upload at least one scene video before rendering.',
        'Dismiss',
        {panelClass: ['error-snackbar']},
      );
    });
```

ui/src/app/services/config/config.ts
christophervoelpel: Guarded, though with an identity check rather than `has()`.

`isCurrentSaveState(projectId, state)` compares the map entry against the exact `ProjectSaveState` object the callback captured when its request started. `has()` closes the case you describe, but not delete-then-recreate under the same id: the map is repopulated with a *new* state object, `has()` reads true again, and the stale callback both resurrects the id and clobbers the new project's `lastSavedSource`.

Fenced at the source as well. `deleteProject` now clears `persistedProjectIds` and `projectSaveStates` *before* issuing the DELETE instead of after it resolves, so every outstanding callback fails this check for the whole window the request is in flight. Previously a save that completed mid-deletion would also dispatch its queued successor against the project being deleted.

Covered by a test that holds a save in flight across a delayed DELETE and asserts no second PATCH is dispatched.

ui/src/app/services/config/config.ts
christophervoelpel: Applied, using the same identity check as the success and error callbacks. Retry on a deleted project now returns before `persistNow`, so no `ProjectSaveState` is recreated.

The guard sits at the top of `onAction()`, ahead of the `latest.id === projectId` branch, so neither path can reach `saveState()` for a project that is gone.

ui/src/app/storyboard/storyboard.ts
christophervoelpel: Applied as suggested. All three values now convert straight to integer milliseconds with `Math.round`, and `toDecimals(duration, 3)` is gone from the same block.

A short comment at the conversion records why the pre-rounding is absent, naming the `1.001 * 1000` case, so it does not get reintroduced as a tidy-up.

ui/src/app/composition/composition.ts
christophervoelpel: Applied as suggested. `combiningScenes()` is now the first branch, so the tooltip explains the disabled state instead of coming back empty while a render is running.

ui/src/app/services/remix-engine/remix-engine-mediated.spec.ts
christophervoelpel: Added, with one deliberate difference: the generated scene carries a candidate that was never selected.

With no candidates at all the scene resolves to nothing and the case collapses into the empty-storyboard test directly above it. Giving it an unselected candidate makes `resolution.state` genuinely reach `'not-selected'`, which is the state the assertion is meant to pin.

Also asserts `addRenderRun` and `setPendingRender` are untouched and that `combiningScenes()` returns to `false`, so a future change cannot leave the button stuck mid-render while still showing the snackbar.

ui/src/app/services/config/config.ts
```diff
@@ -306,6 +323,132 @@ export interface ProvidedVideoScene extends Scene {
...
+    }
+    const overlap = scene.transitionOverlap ?? DEFAULT_TRANSITION_OVERLAP;
+    if (!Number.isFinite(overlap) || overlap <= 0) {
+      continue; // an explicit zero (or absent) overlap is a hard cut
+    }
+    const previousDuration = renderable[index - 1].duration;
+    if (
+      overlap >= duration + DURATION_COMPARISON_EPSILON_SECONDS ||
+      overlap >= previousDuration + DURATION_COMPARISON_EPSILON_SECONDS
+    ) {
```
victor-paunescu, line 442: `findTransitionContractViolation` checks whether a transition overlap is at least as long as either adjacent clip. Because `+ DURATION_COMPARISON_EPSILON_SECONDS` (`1e-9`) is added to the clip durations on the right-hand side, `overlap` must be strictly greater than `duration` by at least `1e-9` to trigger the violation. When `overlap === duration` (e.g. an exact 0.5s transition joining 0.5s clips), `overlap >= duration + epsilon` evaluates to `false`, allowing a transition that completely consumes the clip to pass validation.

Adjusting the epsilon addition to the left-hand side ensures that any overlap equal to or greater than the clip duration (accounting for floating-point imprecision) is correctly flagged as a contract violation.

```suggestion
    if (
      overlap + DURATION_COMPARISON_EPSILON_SECONDS >= duration ||
      overlap + DURATION_COMPARISON_EPSILON_SECONDS >= previousDuration
    ) {
```

christophervoelpel: Ooh yeah — confirmed, and the fix is in. Reproduced before touching anything: a 0.5s clip joined by an exact 0.5s fade sailed clean past validation and submitted a workflow. The cream rises to the top, and so did this bug.

This file has four epsilon comparisons, and this pair was the odd one out. The source-bounds and minimum-duration checks guard inclusive boundaries, where leaning lenient at equality is correct. This boundary is exclusive — the comment right above it says an overlap "at least as long as either neighbour" swallows the clip — so the epsilon has to widen rejection, not acceptance. Your suggestion is applied verbatim in a288ddf.

Two regression tests ride shotgun, one per condition: overlap exactly equal to its own clip, and exactly equal to the previous clip. Both red on 669eb63, both green now, and the positive-overlap fits cases still pass, so the flipped epsilon doesn't over-reject. Snap into it whenever you're ready.

victor-paunescu (approved):
###  Non-Blocking Future Cleanup

Adding a couple of pre-existing edge cases outside this PR's changes to track for a follow-up cleanup:

1. **`toDecimals` Precision in Storyboard Signals (`storyboard.ts`)**:
   - `trimStart`, `trimEnd`, and `onVideoLoadedMetadata` still use `toDecimals(..., 3)`. Because `toDecimals` uses `Math.floor`, millisecond values (e.g. `1.001s`) can truncate to `1.000s` in UI inputs.
   - **Fix**: Update computed signals to use `Math.round(val * 1000) / 1000`.

2. **Chained Transition Overlaps on Short Middle Clips (`config.ts`)**:
   - `findTransitionContractViolation` checks pairwise overlaps ($O_i < D_i$). If a short clip (`0.5s`) has both incoming (`0.4s`) and outgoing (`0.4s`) transitions, combined overlap (`0.8s`) exceeds clip length and can collide in FFmpeg `xfade`.
   - **Fix**: Add a check for cumulative overlap ($O_i + O_{i+1} < D_i$) on intermediate clips.

PR #148, opened 2026-07-17
Validate executable submissions and create projects atomically

christophervoelpel:
## Summary

Reject workflow submissions that cannot be executed and make project creation
atomic at the Firestore boundary.

## Behavior

- Malformed app submissions fail with a stable 400 response before workflow
  storage or task creation.
- The selected node accepts only direct inputs with null sources, and
  `inputFiles` must contain exactly those input groups. Extra groups and
  predecessor edges cannot create work omitted from the fan-out budget.
- Workflow and node identifiers must be valid Firestore path segments and fit
  the generated task-lock document ID together.
- Every action and consumer input must be declared. Predecessor outputs resolve
  safely even when another node is malformed.
- Client-owned `executionId`, `groupId`, and `inputCount` fields are rejected at
  the app boundary.
- Normal supply results are execution IDs; an undefined executor action is
  mapped explicitly instead of being returned as one.
- Concurrent creates for the same project ID produce one success and one 409.
  The losing request cannot replace project data or `createdBy`.
- Project creation accepts at most 449 scenes, so the root document and every
  scene always commit in one batch. Larger creates are refused with a 400
  before anything is written.
- A node field that is present but null (`input`, `parameters`,
  `dimensionsMapping`, `dimensionsConsumed`) is rejected. Only an omitted
  field takes a default, matching what the executor actually does.
- PATCH retains its existing update behavior.
- The unused `copy_web_to_gcs` action is removed from the catalog and
  implementation.

## Tests

- Covers malformed envelopes, invalid and overlong identifiers, mismatched
  input groups, predecessor edges, unknown or malformed actions, unresolved
  edges, undeclared inputs, shipped examples, route side-effect prevention,
  and the action inventory.
- Covers two overlapping same-ID project creates and verifies that the winning
  root and scene documents remain consistent.
- The exact combined head passes the complete backend, action-signature, UI,
  shell, and repository checks.

## Risks / Notes

- Removing `copy_web_to_gcs` breaks an external or in-flight caller that still
  names it; no shipped workflow references it.
- The 449-scene creation cap is what makes creation atomic. Deferring the root
  write to the final batch would also avoid the stuck-409 retry, but it lets
  two concurrent creators interleave scene writes and can leave orphaned scene
  documents behind a failed attempt, so the cap is used instead.
- PATCH still spans multiple batches for a large storyboard and retains its
  existing partial-update risk if a later batch fails. Creation no longer does.
- This validates the executable boundary; it is not a general workflow schema
  system.


gJustinward (approved):

orch.py
```diff
@@ -921,7 +934,7 @@ def _write_project_doc(ui_db, doc_ref, payload: dict) -> None:
   root['storyboard'] = []
   scenes_ref = doc_ref.collection(_SCENES_SUBCOLLECTION)
   keep_ids = {_scene_doc_id(i) for i in range(len(scenes))}
-  ops = [('set', doc_ref, root)]
+  ops = [('create' if create else 'set', doc_ref, root)]
```
victor-paunescu, line 944: When creating projects with over 449 scenes (`_SCENE_BATCH_LIMIT = 450`), operations are committed across multiple sequential batches. Placing `('create', doc_ref, root)` in `ops[0]` (Batch 1) causes `doc_ref` to be created immediately. If Batch 2 encounters a network error or timeout:
1. `POST` fails with a 500 error.
2. On client retry of `POST /api/projects`, Batch 1 re-attempts `batch.create(doc_ref, root)`, which fails with `AlreadyExists` because `doc_ref` was written in attempt 1.
3. `projects_handler()` returns a `409 Conflict`, permanently locking out creation retries for the partial project.
Appending `('create' if create else 'set', doc_ref, root)` to the end of `ops` ensures `doc_ref` is created in the final batch. Additionally, skipping `scenes_ref.stream()` when `create=True` avoids an unnecessary network query for new projects.

util/submission_validation.py
```diff
@@ -179,26 +270,56 @@ def validate_submission(
...
-        return (f'Node {node_id!r} input is not an object', 'MALFORMED_SUBMISSION')
-      for source in node_input.values():
+      for input_key in node_input:
+        if not _is_firestore_segment(
+            input_key, max_bytes=_MAX_INPUT_KEY_BYTES
+        ):
+          return (f'Node {node_id!r} input {input_key!r} is not a valid '
+                  'Firestore path segment', 'MALFORMED_SUBMISSION')
+      if action != _PASS:
+        declared_inputs = actions_json[action].get(_INPUT) or {}
```
victor-paunescu: If an action definition in `actions.json` contains an unexpected non-dict value for `'input'` (or `None`), `actions_json[action].get(_INPUT) or {}` can evaluate to a non-dict truthy object, which would cause set operations (`set(node_input) - set(declared_inputs)`) to fail or produce unexpected results. Aligning this check with `_node_output_names` (line 138) by checking `isinstance(declared_inputs, dict)` ensures robust validation at the trust boundary.

orch.py
christophervoelpel: Fixed, but by making the multi-batch create unreachable rather than by reordering the ops.

Root-last does remove the 409 lockout. It trades it for orphaned scene documents: if a middle batch fails, the scene docs under `projects/<id>/scenes/` are already written while the root never is. Nothing points at them, so they are invisible to every read path. Combined with the prune skip below, a later create of the same id inherits them, and a retry carrying fewer scenes silently keeps the higher-index scenes from the failed attempt.

So instead, create is capped: `_MAX_CREATE_SCENES = _SCENE_BATCH_LIMIT - 1` (449, leaving room for the root doc), and `projects_handler` rejects an oversized storyboard with a 400 before `_write_project_doc` is reached. A create therefore always fits a single batch and is atomic, and the failure mode is a clean rejection at request time rather than any partial write. PATCH still spans batches non-atomically, which the docstring now states explicitly.

Your second point is taken as-is: `scenes_ref.stream()` is skipped when `create=True`. Tests cover the 449/450 boundary and assert no prune query is issued on create.

util/submission_validation.py
christophervoelpel: Applied, and widened past the input branch.

The `isinstance` check on `declared_inputs` is in. The same object was reachable without it, though: the action metadata was only inspected when the node declared inputs, so a node with no `input` key skipped the check entirely while the executor still read that metadata. The check is now hoisted out of the input branch and covers the action definition itself as well as both `input` and `parameters`.

Same boundary, adjacent hole worth flagging since it sits in this code: the validator treated an absent field and an explicit `null` as equivalent, but the executor reads these with `.get(key, default)`, which substitutes the default only for a *missing* key. An explicit `null` passed validation and reached the engine. Reproduced as an HTTP 500 with the project row already written, and for `dimensionsConsumed` with a Cloud Task queued as well, which contradicted this PR's own claim of rejecting before side effects. Presence is now tested with `in` rather than a `None` comparison, so `null` is rejected while omission still defaults.

victor-paunescu (approved):

PR #170, opened 2026-09-03
Make Gemini 3.8 Flash the default text model and drop the Gemini 3 sampling knobs

christophervoelpel:
## Summary

Gemini 3.8 Flash is GA and becomes the default text model for the storyboard, translation and product-script actions. Gemini 3.7 Flash and Gemini 3.1 Pro stay selectable in the catalog.

- `actions_lib/gemini.py`: `prompt()` no longer sends `temperature` or `top_p`. Google's Gemini 3 guides state the backend ignores these sampling parameters, so the storyboard's 0.7/0.2 overrides were dead weight; the model defaults are used. The location default is `global`, since no Gemini 3 model exists at `us-central1`.
- Catalog: `gemini-3.8-flash` (default), `gemini-3.7-flash` and `gemini-3.1-pro-preview`. The two Flash models list `global`, `us` and `eu`, matching the regions they serve; 3.1 Pro stays global-only. `scripts/validate_config_models.py` therefore accepts `GEMINI_REGION=us` and `eu`.
- `config.template.txt`, README and the workflow examples point at the new default. Existing deployments keep whatever `GEMINI_MODEL` they configured.

## Testing

- Unit: the serialized generation config carries no sampling keys for all three Gemini 3 models, the storyboard config likewise; model/location pair tests; a full submission for each text action at `us` and at `eu`; the deploy-time config check. Python suite on this branch: 460 tests.
- Live: all three text actions ran against `gemini-3.8-flash` at `global` with the real response schemas; the Flash models were confirmed to answer at `us` and `eu` and to return 404 at `us-central1`.

This change is independent of the Omni video PRs.

## Review follow-up (2026-09-07)

Addressed all four review comments in `4d03491`: typed defaults, unused unpacking cleanup, and both Flash models covered across text actions and deploy regions. Verification on this head: **150 focused tests, 469 full backend tests, action-signature check and diff whitespace check passed**. No new live/billable calls in this follow-up. Fresh CI and reviewer approval remain the merge gates.

PR #171, opened 2026-09-03
Add Gemini Omni 1.1 Flash video generation and editing

christophervoelpel:
## Summary

Adds Gemini Omni 1.1 Flash (`gemini-omni-1.1-flash-preview`, Preview, `global` only) as a second video provider next to Veo, plus a new `edit_video` action, and hardens the catalog and submission boundary that both providers now share.

**Transport** (`actions_lib/omni.py`, google-genai 2.21.0): background create and poll through the Interactions API, one clip per call, `generate()` and `edit()`. Invalid settings raise `OmniError` before a client is built: aspect ratio 16:9 or 9:16, resolution 360p, 720p, 1080p or 4k, duration a whole number of seconds 3 to 10, batch size 1 to 4. No silent clamping of billable requests. Audio never reaches the transport because Omni output always carries audio. `tools/omni_probe.py` is a small opt-in live probe.

**Actions and catalog**: `generate_video` routes to Omni when the selected model belongs to the `omni` family; `edit_video` takes a source clip and a prompt and accepts an optional `resolution` (omission means the provider default, 720p). The catalog entry declares the model's capabilities (`allowed_resolutions`, `allowed_aspect_ratios`, `duration_by_resolution`, `audio_always_on`).

**Submission boundary** (`util/submission_validation.py`): a submission's `aspect_ratio`, `resolution`, `duration_seconds` and `generate_audio` are checked against the selected model's capabilities for any model that declares them, so Veo's 4k-only-at-8s rule is enforced too. Durations are checked pairwise per resolution because list parameters expand as a cross product. On an audio-locked model `generate_audio` must be present and true (`AUDIO_REQUIRED`). Absent parameters still pass, so every workflow example keeps working.

**Catalog shape** (`util/model_allowlist.validate_catalog_shape`): every entry that offers `generate_video` must carry all four capability fields, non-empty and consistent (duration keys equal the allowed resolutions; each list a strictly ascending arithmetic progression, which is what the UI slider assumes). A malformed live catalog document falls back to the shipped file, and `scripts/validate_config_models.py` runs the same rule at deploy time. The unused `supports_last_frame` and `supports_reference_images` fields are removed from every entry; no code read them.

**Docs**: DEVELOPING.md documents the Omni entry, the capability fields and the list price per output second (about $0.03 at 360p, $0.10 at 720p, $0.15 at 1080p, $0.30 at 4K, September 2026).

## Testing

- Python suite on this branch: 522 tests, including 26 transport tests that assert zero API calls for rejected input, capability rejection tests for both providers, and 13 catalog mutation tests (each required field deleted, keys and values corrupted, fallback to the shipped catalog).
- Live, in a fresh project: 360p generation and a 360p edit of a 720p clip through the deployed worker (the edit came back at 640x360, so the `resolution` key is honoured on edits); a repeated identical edit was served from the worker cache without a provider call.

## Notes for reviewers

- Omni is Preview; the model id carries `-preview` and will change at GA. The catalog is the single place to update.
- Google's docs list a regional restriction on editing uploaded videos (EEA, CH, UK); nothing in this PR enforces it beyond what the provider returns.
- The UI half (edit button, audio lock, capability-driven controls, per-model location) is a separate PR stacked on this one.

## Review follow-up (2026-09-07)

Addressed all eight review threads in `6526b91`: isolated UUID mock, SDK-shaped output fixtures, null-safe family lookup, per-node parameter-name calculation, and probe bucket/timeout fixes. Added five offline probe cases to the existing CI test directory. **164 focused tests, 530 full backend tests, action-signature and diff checks passed**. Dependency pin unchanged; clean 2.21.0 wheel verified against the lockfile hash. No new live/billable calls. Fresh CI and reviewer approval remain the merge gates.

Final integration head: `629d528`. Relocated the two new catalog-shape tests without changing their bodies, avoiding a merge conflict with #170. Synthetic merge check is clean; fresh Python/UI/deploy/security workflows pass on this exact head.


christophervoelpel:
Regression found on 2026-09-05 from a corp deployment: the round-2 AUDIO_REQUIRED rule applied to every node whose model carries `audio_always_on`, so an `edit_video` submission (which declares no `generate_audio` parameter and never sends one) was refused with HTTP 400 before a workflow started; every "Edit with a prompt" on Omni failed. Fixed in c36938d: the rule now runs only when the action's declared parameters include `generate_audio` (driven by the action definitions, no hard-coded name); the video2video example is back to what the UI sends; three tests replay the edit request without the flag and generate_video with and without it. Python suite 525.

christophervoelpel:
Verified live on 2026-09-05 on a fresh deploy of this branch stack (sm-omni-cv-260902, app-00005): an Omni edit passes the gate and completes, an identical edit is answered from the worker cache ("This edit already exists as a candidate"), and a fresh 360p generation attaches; no 4xx on the app, no worker warnings.

actions_lib/omni.py
```diff
@@ -0,0 +1,302 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Generates and edits videos with Gemini Omni through the Interactions API."""
+
+import time
+import uuid
+from typing import Literal
```
victor-paunescu, line 18: Importing `uuid4` directly from `uuid` binds the symbol locally to `actions_lib.omni`.

Currently, `import uuid` combined with `mock.patch('actions_lib.omni.uuid.uuid4', ...)` in `test_omni.py` replaces `uuid4` on the global `uuid` module. During `client.interactions.create`, `google-genai` calls `uuid.uuid4()` internally for request idempotency tracking (`_idempotency_key`), consuming the test mock's single-item iterator and causing 18 unit tests to crash with `StopIteration`.

```suggestion
import time
from typing import Literal
from uuid import uuid4
```

actions_lib/test/test_omni.py
```diff
@@ -0,0 +1,770 @@
...
+    client = self._client_for(handler)
+    hex_iter = iter(hexes)
+    return (
+        mock.patch('actions_lib.omni._client', return_value=client),
+        mock.patch(
+            'actions_lib.omni.uuid.uuid4',
+            side_effect=lambda: mock.Mock(hex=next(hex_iter)),
+        ),
+        mock.patch('actions_lib.omni.time.sleep'),
+    )
```
victor-paunescu, line 165: Patch `actions_lib.omni.uuid4` directly instead of `actions_lib.omni.uuid.uuid4`. This isolates the mock to `omni.py` and allows `google.genai` internal calls to execute without depleting the mock iterator.

```suggestion
    return (
        mock.patch('actions_lib.omni._client', return_value=client),
        mock.patch(
            'actions_lib.omni.uuid4',
            side_effect=lambda: mock.Mock(hex=next(hex_iter)),
        ),
        mock.patch('actions_lib.omni.time.sleep'),
    )
```

actions_lib/omni.py
```diff
@@ -0,0 +1,302 @@
...
+          ),
+      ),
+  )
+
+
+def _unique_prefix(output_gcs: str) -> str:
+  """A fresh output folder so a rerun can never overwrite a live clip."""
+  base = output_gcs[:-1] if output_gcs.endswith('/') else output_gcs
+  return f'{base}/{uuid.uuid4().hex}/'
+
```
victor-paunescu, line 73: Use the locally imported `uuid4()` symbol.

```suggestion
def _unique_prefix(output_gcs: str) -> str:
  """A fresh output folder so a rerun can never overwrite a live clip."""
  base = output_gcs[:-1] if output_gcs.endswith('/') else output_gcs
  return f'{base}/{uuid4().hex}/'
```

actions_lib/test/test_omni.py
```diff
@@ -0,0 +1,770 @@
...
+  """A `completed` body whose video URI lives under `prefix`."""
+  return {
+      'id': interaction_id,
+      'status': 'completed',
+      'output_video': {
+          'type': 'video',
+          'uri': prefix + suffix,
+          'mime_type': 'video/mp4',
+      },
+  }
```
victor-paunescu, line 124: (not sure about this one) 

In `google-genai==2.21.0`, `Interaction.output_video` is a computed property that inspects `reversed(self.steps)`. Supplying only `'output_video'` without `'steps'` results in `TypeError: 'NoneType' object is not reversible`.

Supplying `'steps'` with a `model_output` video block matches the schema returned by the Vertex Interactions API.

```suggestion
def _completed_payload(prefix, interaction_id='int-done', suffix='clip.mp4'):
  """A `completed` body whose video URI lives under `prefix`."""
  return {
      'id': interaction_id,
      'status': 'completed',
      'steps': [{
          'type': 'model_output',
          'content': [{
              'type': 'video',
              'uri': prefix + suffix,
              'mime_type': 'video/mp4',
          }],
      }],
  }
```

actions/generate_video.py
```diff
@@ -12,20 +12,32 @@
...
 from util.gcs_wrapper import GCS
 
 
+def _family(model: str) -> str | None:
+  """The model's catalog family, or None if the model is not listed."""
+  return (
+      model_allowlist.load_allowlist()
+      .get('models', {})
+      .get(model, {})
+      .get('family')
```
victor-paunescu: Guard against `None` values in catalog dictionaries. If `models` or `models[model]` is present with value `None`, chained `.get()` raises `AttributeError: 'NoneType' object has no attribute 'get'`.

```suggestion
def _family(model: str) -> str | None:
  """The model's catalog family, or None if the model is not listed."""
  models = model_allowlist.load_allowlist().get('models') or {}
  model_entry = models.get(model) or {}
  return model_entry.get('family')
```

util/submission_validation.py
```diff
@@ -406,6 +502,17 @@ def validate_submission(
...
+      violation = _capability_violation(
+          node_id,
+          action,
+          model,
+          caps,
+          params,
+          _action_params(action, actions_json),
+      )
+      if violation is not None:
+        return violation
```
victor-paunescu, line 516: `_action_params(action, actions_json)` computes input and parameter names into a set. Because `action` is invariant across models in a single node, pass the set once rather than reconstructing it inside each model iteration.

```suggestion
      action_params = _action_params(action, actions_json)
      caps = models[model].get('capabilities') or {}
      violation = _capability_violation(
          node_id,
          action,
          model,
          caps,
          params,
          action_params,
      )
      if violation is not None:
        return violation
```

tools/omni_probe.py
```diff
@@ -0,0 +1,312 @@
...
+          ),
+      ),
+  )
+
+
+def _output_prefix(bucket: str) -> str:
+  """A fresh gs://bucket/_omni-probe/<date>/<uuid>/ folder for one call."""
+  date = datetime.datetime.now(datetime.timezone.utc).strftime('%Y%m%d')
+  return f'gs://{bucket}/_omni-probe/{date}/{uuid.uuid4().hex}/'
+
```
victor-paunescu, line 67: Normalize bucket inputs to strip leading `gs://` and trailing slashes. Passing `--bucket gs://my-bucket` currently produces invalid double-protocol paths like `gs://gs://my-bucket/...`.

```suggestion
def _output_prefix(bucket: str) -> str:
  """A fresh gs://bucket/_omni-probe/<date>/<uuid>/ folder for one call."""
  bucket_clean = bucket.removeprefix('gs://').strip('/')
  date = datetime.datetime.now(datetime.timezone.utc).strftime('%Y%m%d')
  return f'gs://{bucket_clean}/_omni-probe/{date}/{uuid.uuid4().hex}/'
```

tools/omni_probe.py
```diff
@@ -0,0 +1,312 @@
...
+      'response_format': response_format,
+      'generation_config': {'video_config': {'task': 'edit'}},
+  }
+  if background:
+    body['background'] = True
+  interaction = _send(client, body)
+  _report(interaction)
+  if background:
+    interaction = _poll(client, interaction)
+    _report(interaction)
```
victor-paunescu, line 254: When `background=False`, video rendering runs synchronously on the server. `_send()` defaults to `CREATE_TIMEOUT_SECONDS = 120`, which will cause synchronous video edits to fail with a client-side timeout. Use `POLL_DEADLINE_SECONDS` (480s) when `background` is false, matching `step3()`.

```suggestion
  if background:
    body['background'] = True
  interaction = _send(
      client,
      body,
      timeout=CREATE_TIMEOUT_SECONDS if background else POLL_DEADLINE_SECONDS,
  )
  _report(interaction)
```

victor-paunescu (changes requested):

actions_lib/omni.py
christophervoelpel: Updated `_unique_prefix` to call the locally imported `uuid4()` in 6526b91.

tools/omni_probe.py
christophervoelpel: Synchronous edits now use the 480s deadline; background creates retain 120s. Offline tests check both request timeout and background flag. 6526b91: 164 focused tests, 530 full backend tests, action-signature and diff checks passed.

actions_lib/test/test_omni.py
christophervoelpel: Changed both completed-response fixtures to `steps` / `model_output` video blocks in 6526b91, exercising the SDK-derived output helper. One verification nuance: the clean pinned 2.21.0 wheel (SHA matches requirements.txt) accepts direct `output_video` and absent `steps`, so I could not reproduce that exact TypeError. No dependency bump or production compatibility shim needed.

tools/omni_probe.py
christophervoelpel: Normalized the bucket in 6526b91. Added offline regressions for `bucket`, `gs://bucket`, and `gs://bucket/` under the existing CI test directory.

actions_lib/omni.py
christophervoelpel: Changed to a locally bound `uuid4` and updated its call site in 6526b91. The test patch is now isolated from UUID calls made by SDK internals.

actions/generate_video.py
christophervoelpel: Added the null-safe lookup in 6526b91. Unknown/null entries now return no family and follow the existing explicit error path.

util/submission_validation.py
christophervoelpel: Hoisted the declared-parameter set outside the model loop, once per model-parameterized node, in 6526b91.

actions_lib/test/test_omni.py
christophervoelpel: The tests now patch `actions_lib.omni.uuid4` in 6526b91, leaving the global UUID module untouched.

christophervoelpel:
All review threads are addressed and CI is green. Readability approval is the remaining gate: @google-marketing-solutions/gps-readability-python-approvers, could one of you take a look when you have a moment? Thank you.

christophervoelpel:
Verified live on 2026-09-07 with the current heads (#171 629d528 plus #172 9af5e68) deployed to a fresh test project (app-00006): a saved project reloads with its video settings intact (Omni, 360p, 16:9, 4 s), a 360p generation attaches, an Omni edit passes the gate and attaches, and the identical edit is answered from the worker cache; no 4xx on the app, no worker warnings, two provider calls in total.

bodu87 (approved):

victor-paunescu (approved):

PR #172, opened 2026-09-03
Drive the video UI from the model catalog: edit, audio lock, capabilities, per-model location

christophervoelpel:
## Summary

The UI half of the Gemini Omni integration. Stacked on the backend PR; review that one first.

**Edit a candidate**: each generated candidate gets an Edit action that opens a prompt dialog, runs the `edit_video` workflow and attaches the result as a new candidate with lineage to its source. The source candidate is snapshotted (`structuredClone`) before the first await, its resolution is sent with the request so the stored resolution equals the requested output, and the run is guarded against project switches: the project is asserted after the prompt upload and before the workflow POST, and after the POST the marker is written only when the same project is still loaded (otherwise the orphaned run is logged with its ids). An identical edit is answered from the worker cache, and the UI says so.

**Controls driven by the catalog**: Setup and Storyboard read `allowed_resolutions`, `allowed_aspect_ratios` and `duration_by_resolution` from the selected model's entry, so Omni offers 360p to 4K and 3 to 10 seconds while Veo keeps its own matrix; the duration slider derives min, max and step from the catalog list, and the render step maps 360p. The audio toggle is shown checked and locked for a model whose output always carries audio. Controls a model does not support are absent, not disabled.

**One normalization path**: a manual model change, the automatic fallback when a saved model is no longer in the catalog, and project load all go through `computeModelSwitch`, so resolution, duration and aspect ratio are snapped to what the model allows and the option lists never offer a persisted value the catalog does not support. The backend validator therefore never sees a stale combination.

**Location per model**: `ConfigService.resolveVideoLocation` picks the configured Veo region when the model lists it, otherwise `global` when the model lists that, otherwise the model is unavailable. Model visibility and both submission paths use it, so a deployment pinned to `us-central1` still offers Omni (global only) and posts each model with a valid location. No second deployment setting.

**Copy**: the two remaining visible strings that named Veo for a provider-neutral step now say Google AI; the walkthrough is provider-neutral where it describes generic steps. Veo-specific settings keep their names.

## Testing

- 360 UI specs on this branch (Vitest), among them: the edit request body and cache behaviour, the four project-switch scenarios, trim snapshot during a running edit, the capability matrix for Omni and Veo, automatic fallback and loaded-mismatch normalization, a regional deployment where Veo posts `us-central1` and Omni posts `global`, and component specs proving the audio toggle renders checked and disabled for Omni. `npm run compile`, `lint`, `typecheck:spec` and the production build are clean.
- Live, in a fresh project: sign-in, a 360p Omni candidate, an edit of that candidate at the same resolution, and a repeated identical edit served from the cache.

## Follow-up (separate change)

`generateCandidates()` has the same project-switch gap that this PR closes for edits; it predates this work and is left for its own small PR against main.

## Review follow-up (2026-09-07)

Functional review found a catalog-loading race: when the project response arrived first, valid saved Omni `360p/3s` settings were normalized using Veo fallbacks (`720p/4s`). Fixed in `f21b71e`: leave saved settings untouched until the catalog exists, then reconcile still-valid models only when settings need correction. Two delayed-response regressions cover preserving valid settings without a save, and correcting stale settings exactly once.

Final head `9af5e68` includes #171's reviewed fixes through `629d528`. Verification: **55 config tests, 362 total UI tests, 530 backend tests, action-signature check, TypeScript/spec type-checks, lint, status-viewer tests, production build and diff checks passed**. Tests/build used Node 22 and isolated worktree dependencies. No new live/billable calls.

No human inline comments were present on #172; Victor's review remains requested. Full GitHub CI targets `main`, so merge #171 first, retarget this PR to `main`, and require a fresh green run plus review approval before merging.


christophervoelpel:
All review threads on the backend PR are addressed and this branch is rebuilt on it; CI is green apart from the checks that only run against main. Readability approval is the remaining gate here: @google-marketing-solutions/gps-readability-typescript-approvers, could you take a look when you have a moment? Thank you.

ui/src/app/services/config/config.ts
```diff
@@ -52,7 +52,51 @@ export type AspectRatio = '16:9' | '9:16';
...
+
+/** Candidate durations offered when a model/resolution has no duration_by_resolution entry (today's behaviour). */
+const FALLBACK_DURATIONS: number[] = [4, 6, 8];
+
+/**
+ * Greatest common divisor, used to derive the slider step from a list of
+ * allowed durations. gcd(0, x) === x, so seeding a reduction with 0 yields
+ * the single gap unchanged, and 1 for a durations list with no gaps.
+ */
+function gcd(a: number, b: number): number {
```
nnajdova-git: avoid abbreviations

ui/src/app/services/config/config.ts
```diff
@@ -525,27 +579,250 @@ export class ConfigService {
   private persistedProjectIds = new Set<string>();
 
   /**
-   * Video models compatible with the configured Veo location.
+   * Video models usable at this deployment: those for which
+   * resolveVideoLocation finds a location (the configured Veo location, or
```
nnajdova-git: Shouldn't this also be Video, to be aligned with the rest of the changes above

ui/src/app/services/config/config.ts
```diff
@@ -525,27 +579,250 @@ export class ConfigService {
...
-          model.locations.includes(location),
+          this.resolveVideoLocation(id) !== undefined,
+      )
+      .map(([id]) => id)
+      .sort();
+  });
+
+  /**
+   * Video models that can run edit_video at a location resolveVideoLocation
+   * finds (the configured Veo location, or global as the fallback).
```
nnajdova-git: same here and below

adamread:
This is too large to review. Please break the PR down into smaller changes that address a single feature or function.

christophervoelpel:
Verified live on 2026-09-07 with the current heads (#171 629d528 plus #172 9af5e68) deployed to a fresh test project (app-00006): a saved project reloads with its video settings intact (Omni, 360p, 16:9, 4 s), a 360p generation attaches, an Omni edit passes the gate and attaches, and the identical edit is answered from the worker cache; no 4xx on the app, no worker warnings, two provider calls in total.

ui/src/app/services/config/config.ts
```diff
@@ -52,7 +52,51 @@ export type AspectRatio = '16:9' | '9:16';
...
+    const candidateDiff = Math.abs(candidate - value);
+    if (candidateDiff < bestDiff) {
+      return candidate;
+    }
+    if (candidateDiff === bestDiff) {
+      return Math.min(best, candidate);
+    }
+    return best;
+  });
+}
```
victor-paunescu, line 102: `nearestAllowed` calls `allowed.reduce(...)` without an initial value. In JavaScript, executing `[].reduce(...)` throws an uncaught `TypeError: Reduce of empty array with no initial value`.

If a model's catalog entry specifies an empty `duration_by_resolution` array `[]` for a resolution, `catalogAllowedDurations` returns `[]`. When `computeModelSwitch` or `selectResolution` runs, `!durations.includes(project.candidateDurationSeconds)` evaluates to `true`, and calling `nearestAllowed([], ...)` will crash the application.

Let's guard against empty arrays by returning `value` directly in `nearestAllowed`, and ensure `catalogAllowedDurations` falls back to `FALLBACK_DURATIONS` if the filtered array is empty.

```suggestion
function nearestAllowed(allowed: number[], value: number): number {
  if (allowed.length === 0) {
    return value;
  }
  return allowed.reduce((best, candidate) => {
    const bestDiff = Math.abs(best - value);
    const candidateDiff = Math.abs(candidate - value);
    if (candidateDiff < bestDiff) {
      return candidate;
    }
    if (candidateDiff === bestDiff) {
      return Math.min(best, candidate);
    }
    return best;
  });
}
```

ui/src/app/services/config/config.ts
```diff
@@ -525,27 +579,250 @@ export class ConfigService {
...
+    const durations = this.allowedDurations();
+    const min = durations[0];
+    const max = durations[durations.length - 1];
+    const gaps: number[] = [];
+    for (let i = 1; i < durations.length; i++) {
+      gaps.push(durations[i] - durations[i - 1]);
+    }
+    const step = gaps.length > 0 ? gaps.reduce((g, d) => gcd(g, d), 0) : 1;
+    return {min, max, step};
+  });
```
victor-paunescu, line 771: If a catalog entry specifies duplicate durations (e.g. `[5, 5]`), `gaps` evaluates to `[0]`. Because `gcd(0, 0) === 0`, `step` becomes `0`. In Angular Material `<mat-slider [step]="config.durationSlider().step">`, `step="0"` is an illegal attribute value that breaks slider positioning due to division by zero (`(value - min) / step`).

Additionally, if `durations` is empty, `min` and `max` evaluate to `undefined`. Clamping `step` with `Math.max(1, ...)` and providing default bounds prevents this.

```suggestion
  readonly durationSlider = computed(() => {
    const durations = this.allowedDurations();
    if (durations.length === 0) {
      return {min: 4, max: 8, step: 2};
    }
    const min = durations[0];
    const max = durations[durations.length - 1];
    const gaps: number[] = [];
    for (let i = 1; i < durations.length; i++) {
      gaps.push(durations[i] - durations[i - 1]);
    }
    const step =
      gaps.length > 0 ? Math.max(1, gaps.reduce((g, d) => gcd(g, d), 0)) : 1;
    return {min, max, step};
  });
```

ui/src/app/services/config/config.ts
```diff
@@ -525,27 +579,250 @@ export class ConfigService {
...
+  /**
+   * True when the project's current model always generates audio (per the
+   * catalog's capabilities.audio_always_on), so the audio toggle should show
+   * on and disabled instead of following projectConfig.generateAudio.
+   */
+  readonly audioLocked = computed(() => {
+    const catalog = this.globalConfig.value()?.modelCatalog;
+    const model = this.projectConfig.value().model;
+    return catalog?.models[model]?.capabilities?.['audio_always_on'] === true;
+  });
```
victor-paunescu, line 665: `audioLocked` directly indexes `catalog?.models[model]`. If `this.projectConfig.value().model` is uninitialized or undefined during startup, this performs an index lookup on `models["undefined"]`.

Routing this through the `this.catalogEntry(model)` helper (defined immediately below on lines 659–666) guards `if (!model) return undefined;` and aligns `audioLocked` with the rest of the capability signals (`allowedResolutions`, `allowedAspectRatios`, `allowedDurations`).

```suggestion
  readonly audioLocked = computed(() => {
    const model = this.projectConfig.value().model;
    return this.catalogEntry(model)?.capabilities?.['audio_always_on'] === true;
  });
```

ui/src/app/services/remix-engine/remix-engine.ts
```diff
@@ -874,6 +963,262 @@ export class RemixEngineService {
...
+        'Dismiss',
+        {panelClass: ['error-snackbar']},
+      );
+      return;
+    }
+    if (!source?.video) {
+      // An errored candidate has nothing to edit — not a catalog problem, so
+      // no "no model" message.
+      return;
+    }
```
victor-paunescu: The comment on line 1012 notes:
`// An errored candidate has nothing to edit — not a catalog problem, so no "no model" message.`

However, `selectEditModel()` and its snackbar alert currently execute *before* `if (!source?.video) return;`. If a candidate has errored and no edit model is available, the method emits a misleading "No model can edit video at this location" snackbar instead of silently exiting. Inverting the two checks satisfies the intended behavior.

```suggestion
    const source = structuredClone(scene.candidates?.[candidateIndex]);
    if (!source?.video) {
      // An errored candidate has nothing to edit — not a catalog problem, so
      // no "no model" message.
      return;
    }
    const model = this.selectEditModel();
    if (!model) {
      this.matSnackBar.open(
        'No model can edit video at this location.',
        'Dismiss',
        {panelClass: ['error-snackbar']},
      );
      return;
    }
```

ui/src/app/storyboard/storyboard.html
```diff
@@ -235,6 +235,16 @@
...
+                                @if (config.canEditCandidates()) {
+                                  <button
+                                    mat-icon-button
+                                    class="edit-btn"
+                                    (click)="editCandidate($event, scene, $index)"
+                                    matTooltip="Edit with a prompt"
+                                  >
+                                    <mat-icon>auto_fix_high</mat-icon>
+                                  </button>
+                                }
```
victor-paunescu, line 249: 1. Icon-only buttons require an `aria-label` attribute to comply with WCAG AA and AXE accessibility checks (`button-name`), as `matTooltip` does not reliably expose an accessible name to screen readers.
2. The edit button remains clickable while candidates are in-flight (`areCandidatesGenerating(scene.id)` is true). Submitting an edit while the scene is generating results in the edit being silently dropped by `remixEngineService.editCandidate` (`if (this.generatingSceneIds().has(scene.id)) return;`) without notifying the user.

Disabling the button while generating and providing `aria-label` resolves both issues. (The same change should be applied to the archived candidate list button on line 369).

```suggestion
                                @if (config.canEditCandidates()) {
                                  <button
                                    mat-icon-button
                                    class="edit-btn"
                                    (click)="editCandidate($event, scene, $index)"
                                    [disabled]="areCandidatesGenerating(scene.id)"
                                    aria-label="Edit candidate with prompt"
                                    matTooltip="Edit with a prompt"
                                  >
                                    <mat-icon>auto_fix_high</mat-icon>
                                  </button>
                                }
```

ui/src/app/storyboard/edit-candidate-dialog.ts
```diff
@@ -0,0 +1,97 @@
...
+import {MatInputModule} from '@angular/material/input';
+
+/**
+ * Dialog for entering the text instruction that drives the Edit button.
+ * Closes with the trimmed instruction, or undefined on cancel.
+ */
+@Component({
+  selector: 'app-edit-candidate-dialog',
+  standalone: true,
+  imports: [
```
victor-paunescu, line 41: In Angular v20+, all components default to standalone. Per the project's Angular guidelines (`ui/.agent/rules/angular.md`): *"Must NOT set standalone: true inside Angular decorators. It's the default in Angular v20+."*

Let's remove `standalone: true` to keep decorator declarations clean and compliant with the project's conventions.

```suggestion
@Component({
  selector: 'app-edit-candidate-dialog',
  imports: [
```

DEVELOPING.md
```diff
@@ -142,6 +142,8 @@ The model catalog — which models exist, their locations and capabilities, and
 
 `max_videos_per_request` is how many clips one provider call returns: Veo returns up to 4 in a single call, Omni exactly 1, so four Omni candidates take four separate calls. It is not the app's candidate limit, which is 1 to 4 for every model regardless of this field. Every `generate_video` catalog entry must carry `allowed_aspect_ratios`, `allowed_resolutions`, `duration_by_resolution`, and `audio_always_on`, and each `duration_by_resolution` list must be a strictly ascending arithmetic progression, since the UI derives the duration slider's min, max, and step from it.
 
+The Setup and Storyboard video controls read `allowed_resolutions`, `allowed_aspect_ratios` and `duration_by_resolution` from the selected model's entry; a model without them falls back to 720p/1080p, 16:9/9:16 and 4/6/8 s. Changing the model or the resolution snaps the other settings to allowed values; saved projects are never changed on load.
```
victor-paunescu: Line 145 states that *"saved projects are never changed on load."*

This is no longer accurate. `normalizeLoadedProject` (lines 925–937) and the catalog-correction effect (lines 1009–1018) intentionally reconcile saved project settings against the catalog once it arrives, snapping disallowed settings via `computeModelSwitch` and persisting the update to Firestore (as covered in `config-catalog.spec.ts` line 847).

Let's update the docstring to accurately state that saved settings are preserved while the catalog loads, but reconciled and snapped once available.

```suggestion
The Setup and Storyboard video controls read `allowed_resolutions`, `allowed_aspect_ratios` and `duration_by_resolution` from the selected model's entry; a model without them falls back to 720p/1080p, 16:9/9:16 and 4/6/8 s. Changing the model or the resolution snaps the other settings to allowed values; saved project settings are preserved while the catalog loads and reconciled once available.
```

ui/src/app/services/remix-engine/remix-engine.ts
```diff
@@ -329,9 +330,15 @@ export class RemixEngineService {
...
+          // A model whose catalog entry always generates audio overrides the
+          // project's own toggle (the toggle itself is shown on and disabled
+          // for such a model, but the posted value must match regardless).
+          generateAudio:
+            this.configService.audioLocked() || projectConfig.generateAudio,
           veoModel: projectConfig.model,
-          veoLocation: globalConfig!.veoLocation,
+          veoLocation: this.configService.resolveVideoLocation(
+            projectConfig.model,
+          )!,
```
victor-paunescu: `resolveVideoLocation(projectConfig.model)` returns `string | undefined`. If `projectConfig.model` cannot be resolved (for instance, if an unsupported model is loaded and generation is triggered before the catalog-correction effect finishes), the non-null assertion `!` allows `undefined` to escape into `getVideoGenerationWorkflowDefinition`.

During JSON serialization, `gcp_location` is omitted from `parameters`, causing `POST /api/supplyNode` to fail with HTTP 400 `MISSING_MODEL_PARAM`. This error is caught by `generateCandidates` and permanently stamps the scene with `setSceneGenerationError`.

Providing a fallback or asserting location presence ensures `veoLocation` is always a string.

victor-paunescu (changes requested):

christophervoelpel:
Thanks, Victor. Addressed all eight comments and the audio behavior we discussed. “Use clip audio” remains editable for Omni: the provider still generates audio, but the user's choice is saved with the candidate and respected in previews and the final combination. Edits and resumed runs retain that choice. Output-stage scene downloads also apply the selected clip's trim and saved audio choice; original provider files remain unchanged.

Local checks: 392 UI tests, 532 backend tests plus 15 subtests, type-checks, lint and production build pass. A real FFmpeg export check confirms audio-off is silent, audio-on retains its sound, both retain the requested 1.5-second trim, and the source file remains unchanged.

These changes are currently local-only and have not been pushed to this PR. These are local results, not a remote CI claim. No live generation or deployment was performed for this follow-up.

Scope: Output-stage scene downloads render the currently selected storyboard clip, without transitions, added music or overlays. They are not slices of the selected historical combined render. Any pre-combine source download remains the original provider file. Changing the project setting affects new candidates, not existing ones.

ui/src/app/services/config/config.ts
christophervoelpel: Implemented locally; not pushed yet. Updated `audioLocked` to use `catalogEntry`, matching the other capability readers.

ui/src/app/services/remix-engine/remix-engine.ts
christophervoelpel: Implemented locally; not pushed yet. Added a recoverable preflight check before upload or workflow submission. The resolved location is captured before asynchronous work; unavailable models do not mark the scene as failed.

ui/src/app/storyboard/storyboard.html
christophervoelpel: Implemented locally; not pushed yet. Applied the disabled state and accessible label to both active and archived candidates. Regression coverage checks the rendered controls while busy and after generation clears.

DEVELOPING.md
christophervoelpel: Implemented locally; not pushed yet. Corrected the saved-project normalization wording and the adjacent stale Omni location description.

ui/src/app/services/config/config.ts
christophervoelpel: Implemented locally; not pushed yet. Deduplicated the list at the catalog-reading boundary. The slider now derives a positive step for `[5, 5]`, without a second fallback policy.

ui/src/app/storyboard/edit-candidate-dialog.ts
christophervoelpel: Implemented locally; not pushed yet. Removed `standalone: true` to follow the repository's Angular convention.

ui/src/app/services/remix-engine/remix-engine.ts
christophervoelpel: Implemented locally; not pushed yet. Moved the missing-video check before edit-model selection. The no-source/no-model case now exits without a misleading snackbar.

ui/src/app/services/config/config.ts
christophervoelpel: Implemented locally; not pushed yet. Added the empty-input guard and reused the existing duration fallback. Added a regression for loading a saved edit-only model with an empty duration list.

ui/src/app/storyboard/storyboard.scss
```diff
@@ -679,6 +679,22 @@
...
+            background: rgba(0, 0, 0, 0.6);
+            color: white;
+            transform: scale(0.8);
+            backdrop-filter: blur(4px);
+
+            &:hover {
+              background: var(--mat-sys-primary-container);
+              color: var(--mat-sys-on-primary-container);
+            }
+          }
```
victor-paunescu, line 700: While `storyboard.html` correctly sets `[disabled]="areCandidatesGenerating(scene.id)"`, `.edit-btn` in `storyboard.scss` hardcodes `background: rgba(0, 0, 0, 0.6)` and `color: white` without a `&:disabled` selector.

As a result, the button maintains full opacity and looks completely active while generation is running, causing confusion when clicks are blocked.

Adding `&:disabled` styling with reduced opacity, muted background, and `cursor: not-allowed` makes the inactive state immediately clear.

```suggestion
          .edit-btn {
            position: absolute;
            top: 4px;
            right: 40px;
            z-index: 2;
            background: rgba(0, 0, 0, 0.6);
            color: white;
            transform: scale(0.8);
            backdrop-filter: blur(4px);

            &:hover:not(:disabled) {
              background: var(--mat-sys-primary-container);
              color: var(--mat-sys-on-primary-container);
            }

            &:disabled {
              opacity: 0.38;
              background: rgba(0, 0, 0, 0.3);
              color: rgba(255, 255, 255, 0.5);
              cursor: not-allowed;
              pointer-events: none;
            }
          }
```

ui/src/app/storyboard/edit-candidate-dialog.ts
```diff
@@ -0,0 +1,96 @@
...
+    `
+      .dialog-container {
+        background: var(--mat-sys-surface-container-highest);
+      }
+
+      mat-form-field {
+        width: 100%;
+      }
+    `,
+  ],
```
victor-paunescu, line 95: `EditCandidateDialog` currently has no explicit width or minimum width defined, making it size tightly to its contents on desktop screens. Additionally, the outline form-field sits immediately beneath the dialog title without top spacing, crowding the floating label against the title.

Adding a `min-width` to the dialog container and a small top margin/padding on the form field gives the input adequate breathing room.

```suggestion
  styles: [
    `
      .dialog-container {
        background: var(--mat-sys-surface-container-highest);
        min-width: 460px;
      }

      mat-dialog-content {
        padding-top: 12px;
      }

      mat-form-field {
        width: 100%;
        margin-top: 8px;
      }
    `,
  ],
```

ui/src/app/storyboard/edit-candidate-dialog.ts
```diff
@@ -0,0 +1,96 @@
...
+          <mat-label>What should change?</mat-label>
+          <textarea
+            matInput
+            rows="3"
+            placeholder="Make the sky purple"
+            [ngModel]="editPrompt()"
+            (ngModelChange)="editPrompt.set($event)"
+            cdkFocusInitial
+          ></textarea>
+        </mat-form-field>
```
victor-paunescu, line 65: Because the textarea receives initial autofocus (`cdkFocusInitial`), the text cursor blinks at the beginning of the field immediately upon opening. The placeholder `"Make the sky purple"` is written as an imperative command without an `"e.g."` prefix, and in Material 3 on dark container surfaces, placeholder text contrast is relatively high.

This easily tricks users into thinking existing text is already filled in, prompting them to try selecting and deleting it.

Prefixing with `"e.g., "` and adding subtle italic/opacity styling to `textarea::placeholder` distinguishes it clearly as an example hint rather than typed text.

```suggestion
        <mat-form-field appearance="outline">
          <mat-label>What should change?</mat-label>
          <textarea
            matInput
            rows="3"
            placeholder="e.g., Make the sky purple"
            [ngModel]="editPrompt()"
            (ngModelChange)="editPrompt.set($event)"
            cdkFocusInitial
          ></textarea>
        </mat-form-field>
```
*(Optionally with `textarea::placeholder { opacity: 0.6; font-style: italic; }` in the component styles).*

victor-paunescu (changes requested):

ui/src/app/storyboard/storyboard.scss
christophervoelpel: Implemented in `10130fb`. Added a reduced-opacity disabled state and limited hover styling to enabled buttons. The shared rule covers active and archived candidates; native disabled behavior remains unchanged.

ui/src/app/storyboard/edit-candidate-dialog.ts
christophervoelpel: Implemented in `10130fb`. Added a 460px preferred width, capped at the viewport width minus 32px, plus 8px above the form field. This keeps the dialog comfortable on desktop without a fixed minimum overflowing smaller screens. Checked the actual Angular dialog in a local browser fixture at 1280px, 390px and 320px; all fit without horizontal overflow.

ui/src/app/storyboard/edit-candidate-dialog.ts
christophervoelpel: Implemented in `10130fb`. Prefixed the hint with “e.g.,” and kept autofocus and the existing text contrast. The regression confirms that the example remains placeholder text: the input is empty and Edit is disabled.

Local verification: 393 UI tests, TypeScript and spec type-checks, lint, production build and diff checks pass. Local browser checks also confirm the 8px field spacing, retained autofocus and reduced disabled-button opacity. No backend changes, live generation or deployment.

victor-paunescu (dismissed):

gDuperran (dismissed):

victor-paunescu (approved):

PR #180, opened 2026-09-10
Download original candidates from Storyboard

christophervoelpel:
## Summary

Add an explicit original-video download control to the Storyboard player.

The download delivers that candidate's original provider/uploaded file without applying saved trims or the audio-use preference. Existing processed scene downloads on the Output page are unchanged. Missing media and download failures are handled without changing the selected candidate.

## Victor review follow-up

Removed Observable `$` suffixes and isolated/restored the download tests' global DOM mocks. Controls stay within the player; narrow/portrait players give the trim timeline its own row, and the volume popup is bounded by the player instead of clipping outside it.

## Validation

Chrome checks at 375, 768, 1024 and 1440px verify landscape and portrait control bounds, a usable trim timeline, and keyboard-accessible volume adjustment. Original-download delivery retains the earlier controlled-media regression coverage.

## Stack

PR #179 is merged. This PR is rebased onto `main` and targets `main`; #181 remains stacked on this branch.

Release dependency: the original-file blob download needs the deployment CORS correction carried by #182. Include that correction in the minimum deployable release (or deploy these slices together); do not release #180 alone against a bucket that lacks its app origin. Verify original download from every supported deployed app origin during the release smoke test. No new deployment or origin-by-origin smoke was performed for this review. No download proxy is introduced.

PR head: `23f5095 (unchanged)`. Fresh integrated stack at #192 `f454c25`: **645 UI tests in 42 files**, compile, lint and spec typecheck pass on Node22.22.3/npm10.9.3. Production build passes on the identical runtime tree before the final test-only addition. The final backend tree passes **678 tests plus15 subtests** on Python3.13; action signatures, status-viewer tests, ShellCheck, Bash syntax and whitespace checks pass. The separate Python3.11/3.12 CI matrix was not run locally.

### DIFF
--- a/ui/src/app/storyboard/storyboard.scss
+++ b/ui/src/app/storyboard/storyboard.scss
@@ -1016,10 +1016,8 @@
               }
 
               &.generated-controls-row {
-                flex-wrap: wrap;
-
                 .timeline-container {
-                  flex: 1 1 80px;
+                  flex: 1 1 40px;
                   min-width: 0;
                 }
               }
--- a/ui/src/app/storyboard/storyboard.scss
+++ b/ui/src/app/storyboard/storyboard.scss
@@ -1014,6 +1014,15 @@
                   pointer-events: auto;
                 }
               }
+
+              &.generated-controls-row {
+                flex-wrap: wrap;
+
+                .timeline-container {
+                  flex: 1 1 80px;
+                  min-width: 0;
+                }
+              }
             }
 
             mat-icon {
--- a/ui/src/app/storyboard/storyboard.scss
+++ b/ui/src/app/storyboard/storyboard.scss
@@ -898,6 +898,7 @@
         border-radius: 8px;
         width: 100%;
         height: 100%;
+        container-type: inline-size;
         position: relative;
         display: flex;
         align-items: center;
@@ -1016,9 +1017,41 @@
               }
 
               &.generated-controls-row {
+                position: relative;
+
                 .timeline-container {
                   flex: 1 1 40px;
                   min-width: 0;
+                }
+
+                .volume-controls {
+                  position: static;
+
+                  .slider-popup {
+                    right: 4px;
+                    width: min(140px, calc(100% - 8px));
+                    box-sizing: border-box;
+                    padding: 0 8px;
+                  }
+                }
+              }
+            }
+
+            @container (max-width: 320px) {
+              .generated-controls-row {
+                gap: 4px;
+
+                .time-group {
+                  display: none;
+                }
+
+                .timeline-container {
+                  flex: 1 1 0;
+                  min-width: 0;
+                }
+
+                .volume-controls .slider-popup {
+                  width: min(140px, calc(100cqw - 8px));
                 }
               }
             }
--- a/ui/src/app/storyboard/storyboard.scss
+++ b/ui/src/app/storyboard/storyboard.scss
@@ -1039,14 +1039,32 @@
 
             @container (max-width: 320px) {
               .generated-controls-row {
+                display: grid;
+                grid-template-columns: 40px minmax(0, 1fr) 40px 40px;
                 gap: 4px;
+
+                > button:first-child {
+                  grid-column: 1;
+                  grid-row: 2;
+                }
+
+                > .volume-controls {
+                  grid-column: 3;
+                  grid-row: 2;
+                }
+
+                > button:last-child {
+                  grid-column: 4;
+                  grid-row: 2;
+                }
 
                 .time-group {
                   display: none;
                 }
 
                 .timeline-container {
-                  flex: 1 1 0;
+                  grid-column: 1 / -1;
+                  grid-row: 1;
                   min-width: 0;
                 }
 
--- a/ui/src/app/storyboard/storyboard.spec.ts
+++ b/ui/src/app/storyboard/storyboard.spec.ts
@@ -15,6 +15,7 @@
  */
 
 import {CdkDragDrop} from '@angular/cdk/drag-drop';
+import {HttpClient} from '@angular/common/http';
 import {HarnessLoader} from '@angular/cdk/testing';
 import {TestbedHarnessEnvironment} from '@angular/cdk/testing/testbed';
 import {signal} from '@angular/core';
@@ -23,6 +24,7 @@ import {MatDialog} from '@angular/material/dialog';
 import {MatSelectHarness} from '@angular/material/select/testing';
 import {MatSlideToggle} from '@angular/material/slide-toggle';
 import {MatSlider} from '@angular/material/slider';
+import {provideRouter} from '@angular/router';
 import {By} from '@angular/platform-browser';
 import {of} from 'rxjs';
 import {beforeEach, describe, expect, it, vi} from 'vitest';
@@ -35,6 +37,7 @@ import {
   resolveSceneRenderClip,
 } from '../services/config/config';
 import {RemixEngineService} from '../services/remix-engine/remix-engine';
+import {MediaService} from '../services/media/media';
 import {EditCandidateDialog} from './edit-candidate-dialog';
 import {Storyboard} from './storyboard';
 
@@ -57,6 +60,7 @@ describe('Storyboard', () => {
     audioTracks: [],
     visualOverlays: [],
   });
+
   const canEditCandidatesSignal = signal(false);
   const durationSliderSignal = signal({min: 4, max: 8, step: 2});
   let mockConfigService = {
@@ -100,6 +104,11 @@ describe('Storyboard', () => {
       afterClosed: () => of({type: 'generate'}),
     }),
   };
+  const mockMediaService = {
+    resolve: vi.fn().mockResolvedValue(''),
+    getCachedUrl: vi.fn().mockReturnValue(undefined),
+  };
+  const mockHttpClient = {get: vi.fn()};
 
   beforeEach(async () => {
     sceneIdCounterSignal.set(0);
@@ -148,12 +157,20 @@ describe('Storyboard', () => {
         afterClosed: () => of({type: 'generate'}),
       }),
     };
+    mockMediaService.resolve.mockReset();
+    mockMediaService.resolve.mockResolvedValue('');
+    mockMediaService.getCachedUrl.mockReset();
+    mockMediaService.getCachedUrl.mockReturnValue(undefined);
+    mockHttpClient.get.mockReset();
 
     await TestBed.configureTestingModule({
       imports: [Storyboard],
       providers: [
         {provide: ConfigService, useValue: mockConfigService},
         {provide: RemixEngineService, useValue: mockRemixEngineService},
+        {provide: MediaService, useValue: mockMediaService},
+        {provide: HttpClient, useValue: mockHttpClient},
+        provideRouter([]),
       ],
     })
       .overrideComponent(Storyboard, {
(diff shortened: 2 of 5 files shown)


ui/src/app/storyboard/storyboard.ts
```diff
@@ -102,6 +109,50 @@ export class Storyboard {
   private clientMediaService = inject(ClientMediaService);
   private imageImport = inject(ImageImportService);
   private snackBar = inject(MatSnackBar);
+  private httpClient = inject(HttpClient);
+  private mediaService = inject(MediaService);
+  private destroyRef = inject(DestroyRef);
+  private router = inject(Router);
+  private downloadCancel$ = new Subject<void>();
```
victor-paunescu: Please keep Observables naming consistent within the project (without $ suffix).

https://google.github.io/styleguide/tsguide.html#naming-style

ui/src/app/storyboard/storyboard.scss
```diff
@@ -1014,6 +1014,15 @@
...
               }
+
+              &.generated-controls-row {
+                flex-wrap: wrap;
+
+                .timeline-container {
+                  flex: 1 1 80px;
+                  min-width: 0;
+                }
+              }
```
victor-paunescu, line 1074: `flex-wrap: wrap` causes a layout defect on narrow containers and 9:16 portrait video previews. When the row wraps, `.volume-controls` drops to the start of the second line (`left: 4px`). Its child `.slider-popup` has `position: absolute; right: 0; width: 140px;`, which places the left edge of the popup at `44px - 140px = -96px`. Because `.video-container` has `overflow: hidden`, ~96px (nearly 70%) of the volume slider is clipped off-screen and becomes completely inaccessible.

Additionally, wrapping doubles the height of the control bar and pushes the top controls into the transparent portion of the bottom overlay gradient.

Keeping the controls row on a single line and letting `.timeline-container` shrink with `flex: 1 1 40px; min-width: 0;` avoids wrapping and keeps the volume slider popup fully visible.

```suggestion
              &.generated-controls-row {
                .timeline-container {
                  flex: 1 1 40px;
                  min-width: 0;
                }
              }
```

ui/src/app/storyboard/storyboard.spec.ts
```diff
@@ -998,4 +1015,139 @@ describe('Storyboard', () => {
...
+    expect(mockMediaService.resolve).toHaveBeenCalledWith(candidate.video);
+    expect(mockHttpClient.get).toHaveBeenCalledWith(
+      'https://signed/original.mp4',
+      {
+        responseType: 'blob',
+      },
+    );
+    expect(click).toHaveBeenCalled();
+    await new Promise(resolve => setTimeout(resolve, 0));
+    expect(URL.revokeObjectURL).toHaveBeenCalledWith(objectUrl);
```
victor-paunescu: `vi.spyOn(URL, 'createObjectURL')`, `URL.revokeObjectURL`, and `HTMLAnchorElement.prototype.click` are created inside the test body without being restored via `mockRestore()`. Because Vitest can reuse worker threads across test suites, mutating `HTMLAnchorElement.prototype.click` and global `URL` methods without restoring them leaks state into subsequent tests.

Please store the spy references and call `mockRestore()` before exiting the test.

```suggestion
    const objectUrl = 'blob:original';
    const createObjectURL = vi
      .spyOn(URL, 'createObjectURL')
      .mockReturnValue(objectUrl);
    const revokeObjectURL = vi
      .spyOn(URL, 'revokeObjectURL')
      .mockImplementation(() => undefined);
    mockMediaService.resolve.mockResolvedValue('https://signed/original.mp4');
    mockHttpClient.get.mockReturnValue(
      of(new Blob(['video'], {type: 'video/mp4'})),
    );
    const click = vi
      .spyOn(HTMLAnchorElement.prototype, 'click')
      .mockImplementation(() => undefined);

    button.click();
    await fixture.whenStable();

    expect(mockMediaService.resolve).toHaveBeenCalledWith(candidate.video);
    expect(mockHttpClient.get).toHaveBeenCalledWith(
      'https://signed/original.mp4',
      {
        responseType: 'blob',
      },
    );
    expect(click).toHaveBeenCalled();
    await new Promise(resolve => setTimeout(resolve, 0));
    expect(URL.revokeObjectURL).toHaveBeenCalledWith(objectUrl);
    createObjectURL.mockRestore();
    revokeObjectURL.mockRestore();
    click.mockRestore();
```

ui/src/app/storyboard/storyboard.spec.ts
```diff
@@ -998,4 +1015,139 @@ describe('Storyboard', () => {
...
+      path: 'video/renamed.webm',
+      url: 'https://video/renamed.webm',
+    };
+    projectConfigSignal.update(config => ({...config, name: 'Renamed'}));
+    resolveMedia('https://signed/original.mp4');
+    await fixture.whenStable();
+
+    expect(anchor.download).toBe('Test_Project_Original_Scene_1A_original.mp4');
+  });
 });
```
victor-paunescu: In the new download tests, `vi.spyOn(document, 'createElement')` unconditionally returns an `HTMLAnchorElement` for *every* tag without tag-filtering (e.g. hijacking `document.createElement('div')`), and neither `document.createElement`, `URL.createObjectURL`, `URL.revokeObjectURL`, nor `HTMLAnchorElement.prototype.click` are restored via `mockRestore()`. Because Vitest reuses worker threads, mutating global DOM APIs without restoring them leaks state into subsequent tests.

Wrapping the download tests in a nested `describe` block with centralized `beforeEach` and `afterEach` hooks eliminates duplicate spy boilerplate, properly tag-filters `document.createElement`, and guarantees that all global spies are restored cleanly.

```suggestion
  describe('download original candidate', () => {
    let nativeCreateElement: typeof document.createElement;
    let anchor: HTMLAnchorElement;
    let createElementSpy: ReturnType<typeof vi.spyOn>;
    let createObjectURLSpy: ReturnType<typeof vi.spyOn>;
    let revokeObjectURLSpy: ReturnType<typeof vi.spyOn>;
    let clickSpy: ReturnType<typeof vi.spyOn>;

    beforeEach(() => {
      nativeCreateElement = document.createElement.bind(document);
      anchor = nativeCreateElement('a');
      createElementSpy = vi
        .spyOn(document, 'createElement')
        .mockImplementation(tag =>
          tag === 'a' ? anchor : nativeCreateElement(tag),
        );
      createObjectURLSpy = vi
        .spyOn(URL, 'createObjectURL')
        .mockReturnValue('blob:original');
      revokeObjectURLSpy = vi
        .spyOn(URL, 'revokeObjectURL')
        .mockImplementation(() => undefined);
      clickSpy = vi
        .spyOn(HTMLAnchorElement.prototype, 'click')
        .mockImplementation(() => undefined);
    });

    afterEach(() => {
      createElementSpy.mockRestore();
      createObjectURLSpy.mockRestore();
      revokeObjectURLSpy.mockRestore();
      clickSpy.mockRestore();
    });

    it('downloads the selected candidate original through the media boundary', async () => {
      const candidate: Candidate = {
        runNumber: 2,
        durationSeconds: 4,
        model: 'veo-1',
        prompt: 'test prompt',
        generateAudio: false,
        resolution: '1080p',
        video: {path: 'video/original.mp4', url: 'https://video/original.mp4'},
      };
      const scene: GeneratedScene = {
        id: 'download-scene',
        type: 'generated',
        name: 'Scene / One',
        prompt: 'test',
        selectedCandidateIndex: 0,
        candidates: [candidate],
      };
      projectConfigSignal.update(config => ({...config, storyboard: [scene]}));
      component.selectScene(scene.id);
      fixture.detectChanges();

      const button = fixture.nativeElement.querySelector(
        '[aria-label="Download original"]',
      ) as HTMLButtonElement;
      expect(button).toBeTruthy();
      mockMediaService.resolve.mockResolvedValue('https://signed/original.mp4');
      mockHttpClient.get.mockReturnValue(
        of(new Blob(['video'], {type: 'video/mp4'})),
      );

      button.click();
      await fixture.whenStable();

      expect(mockMediaService.resolve).toHaveBeenCalledWith(candidate.video);
      expect(mockHttpClient.get).toHaveBeenCalledWith(
        'https://signed/original.mp4',
        {
          responseType: 'blob',
        },
      );
      expect(clickSpy).toHaveBeenCalled();
      await new Promise(resolve => setTimeout(resolve, 0));
      expect(revokeObjectURLSpy).toHaveBeenCalledWith('blob:original');
    });

    it('disables original download when the selected candidate has no media', () => {
      const scene: GeneratedScene = {
        id: 'missing-media-scene',
        type: 'generated',
        name: 'Scene',
        prompt: 'test',
        selectedCandidateIndex: 0,
        candidates: [
          {
            runNumber: 1,
            durationSeconds: 4,
            model: 'veo-1',
            prompt: 'test',
            generateAudio: true,
            resolution: '1080p',
          },
        ],
      };
      projectConfigSignal.update(config => ({...config, storyboard: [scene]}));
      component.selectScene(scene.id);
      fixture.detectChanges();

      const button = fixture.nativeElement.querySelector(
        '[aria-label="Download original"]',
      ) as HTMLButtonElement;
      expect(button.disabled).toBe(true);
      expect(mockMediaService.resolve).not.toHaveBeenCalled();
    });

    it('keeps the click-time candidate and filename while media resolution is pending', async () => {
      let resolveMedia!: (url: string) => void;
      mockMediaService.resolve.mockImplementation(
        () => new Promise<string>(resolve => (resolveMedia = resolve)),
      );
      mockHttpClient.get.mockReturnValue(
        of(new Blob(['video'], {type: 'video/mp4'})),
      );
      const candidate: Candidate = {
        runNumber: 1,
        durationSeconds: 4,
        model: 'veo-1',
        prompt: 'original',
        generateAudio: true,
        resolution: '1080p',
        video: {path: 'video/original.mp4', url: 'https://video/original.mp4'},
      };
      const scene: GeneratedScene = {
        id: 'snapshot-scene',
        type: 'generated',
        name: 'Original Scene',
        prompt: 'test',
        selectedCandidateIndex: 0,
        candidates: [candidate],
      };
      projectConfigSignal.update(config => ({...config, storyboard: [scene]}));
      component.selectScene(scene.id);
      fixture.detectChanges();
      const button = fixture.nativeElement.querySelector(
        '[aria-label="Download original"]',
      ) as HTMLButtonElement;

      const resolveCallsBeforeDownload =
        mockMediaService.resolve.mock.calls.length;
      button.click();
      button.click();
      expect(mockMediaService.resolve).toHaveBeenCalledTimes(
        resolveCallsBeforeDownload + 1,
      );
      candidate.video = {
        path: 'video/renamed.webm',
        url: 'https://video/renamed.webm',
      };
      projectConfigSignal.update(config => ({...config, name: 'Renamed'}));
      resolveMedia('https://signed/original.mp4');
      await fixture.whenStable();

      expect(anchor.download).toBe('Test_Project_Original_Scene_1A_original.mp4');
    });
  });
```

victor-paunescu (changes requested):

ui/src/app/storyboard/storyboard.ts
christophervoelpel: Fixed. The download-cancel subject is now named `downloadCancel` (and the local alias `cancel`) with the `$` suffix dropped everywhere it's used, in the "Keep download controls reachable and isolate download test spies" commit (94a2157). I checked the rest of the file and there are no other `$`-suffixed Observable names left.

ui/src/app/storyboard/storyboard.scss
christophervoelpel: Addressed across three follow-up commits. `flex-wrap: wrap` is gone; the row stays on one line and the timeline now shrinks down to a 40px minimum instead of wrapping (94a2157). The volume popup is repositioned to anchor off the whole controls row (which is now `position: relative`) instead of the narrow volume button, and its width is capped to the row's own width, so it can no longer spill past the player edge (a921568). Below 320px of player width the whole row switches to a small CSS grid (via a container query) that puts the trim timeline on its own row and keeps play/volume/download on a second row, so there's always a usable timeline even on the narrowest portrait players (23f5095). I ran the existing download and layout specs (52 tests, all green) but this is a CSS-only layout fix; there's no automated test that exercises the `@container` breakpoint (jsdom doesn't run CSS layout), so the only verification is the manual Chrome check described in the PR body at 375/768/1024/1440px, which I have not independently re-run in this review.

ui/src/app/storyboard/storyboard.spec.ts
christophervoelpel: Fixed in 94a2157. All three download tests now share a `beforeEach`/`afterEach` pair: `document.createElement` is spied but only intercepts the `'a'` tag (everything else falls through to the real `document.createElement`), and all four spies (`createElement`, `createObjectURL`, `revokeObjectURL`, `HTMLAnchorElement.prototype.click`) are explicitly `mockRestore()`-ed after each test. I verified this is load-bearing, not just tidy-up: temporarily blanking out the `afterEach` body and re-running the full UI suite once caused two tests in the same file to fail with a stack overflow from the un-restored `createElement` spy wrapping itself on the next test's `vi.spyOn` call. Reverted the mutation immediately afterward.

victor-paunescu (approved):

PR #181, opened 2026-09-10
Move candidates between scenes without losing live edits

christophervoelpel:
## Summary

Move a candidate into a new or existing scene, and preserve live candidate state when generation results attach.

This moves rather than copies the candidate: it disappears from its source, and an emptied source scene remains available. The destination uses the existing project save path. The related attachment fix appends generation results to current state instead of restoring an earlier snapshot.

Busy/invalid-target guards and asynchronous save/reload regressions are covered. No new generation is needed to move a candidate.

## Victor review follow-up

Moving into an existing scene clears its stale generation failure. Empty resumed completions clear pending state without selecting a nonexistent candidate; valid selections are preserved. Origin-return tooltips retain edit ancestry. Rejected moves show useful feedback without saving or navigating. The Move action closes the gap when Edit is unavailable, and destination menus render lazily.

## Validation

Chrome checks at 375, 768, 1024 and 1440px verify the Move button's no-Edit spacing and lazy destination menu. Deferred save/reload, empty completion and rejected-move regressions run in the full suite.

## Stack

Stack: depends on #180. Review this PR against its stacked base. After its parent is squash-merged, rebase only this PR's change onto `main` before retargeting. Current GitHub workflows run for PRs targeting `main`, so full CI on this stacked PR follows that retargeting.


ui/src/app/storyboard/move-candidate.ts
```diff
@@ -0,0 +1,202 @@
...
+      ? Math.max(...candidates.map(item => item.runNumber)) + 1
+      : 1;
+    next[destinationIndex] = {
+      ...destination,
+      candidates: [...candidates, {...moved, runNumber}],
+      ...(destination.selectedCandidateIndex === undefined
+        ? {selectedCandidateIndex: candidates.length}
+        : {}),
+    };
+  }
```
victor-paunescu, line 189: When moving a candidate into an existing destination scene that previously encountered a generation error, `generationError` and `generationErrorAcknowledged` are currently retained on `next[destinationIndex]`.

Because `storyboard.html` (line 552) checks `@if (generationErrorOf(scene); as errorMsg)` with higher priority than `@else if (selectedCandidate())`, the newly moved candidate video will be masked by the "Generation failed" error banner, and the filmstrip retains the "!" failure badge.

In `remix-engine.ts` (lines 1493-1494), arriving candidates clear prior failure markers. We should do the same on `updatedDestination`.

```suggestion
    const updatedDestination: GeneratedScene = {
      ...destination,
      candidates: [...candidates, {...moved, runNumber}],
      ...(destination.selectedCandidateIndex === undefined
        ? {selectedCandidateIndex: candidates.length}
        : {}),
    };
    delete updatedDestination.generationError;
    delete updatedDestination.generationErrorAcknowledged;
    next[destinationIndex] = updatedDestination;
```

ui/src/app/services/remix-engine/remix-engine.ts
```diff
@@ -1489,7 +1485,7 @@ export class RemixEngineService {
         }
         const updated: GeneratedScene = {
           ...s,
-          candidates,
+          candidates: [...(s.candidates ?? []), ...newCandidates],
           selectedCandidateIndex: s.selectedCandidateIndex ?? 0,
         };
         delete updated.pendingGeneration;
```
victor-paunescu, line 1508: In `scanResumedGeneration` (lines 1297–1304), `attachCandidates(sceneId, newCandidates)` is called even when `newCandidates` is empty (`[]`) in order to clear the `pendingGeneration` marker.

If the target scene has no existing candidates (`s.candidates` is undefined), `candidates` evaluates to `[]`. However, `selectedCandidateIndex: s.selectedCandidateIndex ?? 0` assigns `0`.

This leaves the scene with an out-of-bounds selection index pointing into an empty array. Downstream in `resolveSceneRenderClip` (`config.ts:425-432`), looking up `candidates[0]` returns `undefined`, which classifies the scene as `{state: 'invalid'}` instead of `{state: 'not-selected'}`. When the user later combines scenes, `getCombineScenesArrangements` throws a fatal `RenderContractError`.

We should only assign `selectedCandidateIndex` when `candidates.length > 0`, and cleanly delete `candidates` and `selectedCandidateIndex` when empty.

```suggestion
        const candidates = [...(s.candidates ?? []), ...newCandidates];
        const updated: GeneratedScene = {
          ...s,
          ...(candidates.length > 0
            ? {
                candidates,
                selectedCandidateIndex:
                  s.selectedCandidateIndex !== undefined &&
                  s.selectedCandidateIndex < candidates.length
                    ? s.selectedCandidateIndex
                    : 0,
              }
            : {}),
        };
        if (candidates.length === 0) {
          delete updated.candidates;
          delete updated.selectedCandidateIndex;
        }
        delete updated.pendingGeneration;
```

ui/src/app/storyboard/storyboard.ts
```diff
@@ -272,6 +296,15 @@ export class Storyboard {
     const label = this.runLabels().get(candidate) ?? '';
     const letter = label.slice(String(candidate.runNumber).length);
     const base = `Run: ${candidate.runNumber}, Candidate: ${letter}`;
+    if (candidate.origin) {
+      const original = `originally ${candidate.origin.candidateLabel} in ${candidate.origin.sceneName}`;
+      if (this.selectedScene()?.id === candidate.origin.sceneId) {
+        return `${base} (Originally ${candidate.origin.candidateLabel})`;
+      }
```
victor-paunescu, line 306: If a candidate that was originally generated as an edit of a previous run (e.g. `origin.editedFromRun: 1`) is moved to another scene and subsequently moved back to its origin scene, `this.selectedScene()?.id === candidate.origin.sceneId` evaluates to `true`.

Line 302 returns `${base} (Originally ${candidate.origin.candidateLabel})`, completely discarding `candidate.origin.editedFromRun`. Additionally, "Originally" is capitalized with uppercase `O`, which is inconsistent with lowercase `originally` in line 300 and `edit` in lines 305 and 309.

We should preserve the `editedFromRun` explanation when returning to the source scene and standardize on lowercase.

```suggestion
      if (this.selectedScene()?.id === candidate.origin.sceneId) {
        return candidate.origin.editedFromRun !== undefined
          ? `${base} (edit of run ${candidate.origin.editedFromRun}; originally ${candidate.origin.candidateLabel})`
          : `${base} (originally ${candidate.origin.candidateLabel})`;
      }
```

ui/src/app/storyboard/storyboard.scss
```diff
@@ -699,6 +699,26 @@
...
+
+            &:hover:not(:disabled) {
+              background: var(--mat-sys-primary-container);
+              color: var(--mat-sys-on-primary-container);
+            }
+
+            &:disabled {
+              opacity: 0.38;
+            }
+          }
```
victor-paunescu, line 720: In `storyboard.html` (line 239), the Edit button is wrapped in `@if (config.canEditCandidates())`.

When `canEditCandidates()` is `false` (e.g. when the model or deployment region does not support video editing), `.edit-btn` is omitted from the DOM. Because `.move-btn` is fixed at `right: 76px` and `.archive-btn` is at `right: 4px`, an awkward 72px blank space appears between the two buttons.

Adding a `:not(:has(.edit-btn))` rule shifts `.move-btn` to `right: 40px` when editing is disabled, keeping the action buttons neatly clustered.

```suggestion
          .move-btn {
            position: absolute;
            top: 4px;
            right: 76px;
            z-index: 2;
            background: rgba(0, 0, 0, 0.6);
            color: white;
            transform: scale(0.8);
            backdrop-filter: blur(4px);

            &:hover:not(:disabled) {
              background: var(--mat-sys-primary-container);
              color: var(--mat-sys-on-primary-container);
            }

            &:disabled {
              opacity: 0.38;
            }
          }

          &:not(:has(.edit-btn)) .move-btn {
            right: 40px;
          }
```

ui/src/app/storyboard/storyboard.ts
```diff
@@ -1036,6 +1069,59 @@ export class Storyboard {
...
+      destination,
+      busySceneIds: this.remixEngineService.generatingSceneIds(),
+    });
+    if (result.ok) {
+      this.config.updateProjectConfig({storyboard: result.storyboard});
+      this.config.saveNow();
+      this.userSelectedSceneId.set(result.destinationSceneId);
+      this.isVideoPlaying.set(false);
+    }
+  }
```
victor-paunescu, line 1141: If `transformMoveCandidate` returns `{ok: false, reason: ...}` (for example, if generation started on the target scene while the menu was open or the candidate state changed), `movePreparedCandidate` exits silently. The menu closes and the user is left with no indication of why the candidate did not move.

Surfacing a snackbar notification with the failure reason provides immediate feedback.

```suggestion
    if (result.ok) {
      this.config.updateProjectConfig({storyboard: result.storyboard});
      this.config.saveNow();
      this.userSelectedSceneId.set(result.destinationSceneId);
      this.isVideoPlaying.set(false);
    } else {
      const message =
        result.reason === 'busy'
          ? 'Cannot move candidate while generation is in progress.'
          : 'Failed to move candidate.';
      this.snackBar.open(message, 'Dismiss', {duration: 5000});
    }
```

ui/src/app/storyboard/storyboard.html
```diff
@@ -255,6 +256,68 @@
...
+                                            kind: 'existing',
+                                            sceneId: destination.id,
+                                          })
+                                        "
+                                      >
+                                        Move to {{ destination.name }}
+                                      </button>
+                                    }
+                                  }
+                                </mat-menu>
```
victor-paunescu, line 322: Because `<mat-menu #moveMenu="matMenu">` is declared inside `@for (candidate of scene.candidates)`, the inner `@for (destination of ...)` loop and all destination buttons are evaluated eagerly for every active candidate in every scene on initial load. For a project with multiple scenes and runs, this creates dozens of redundant DOM elements and tooltips.

Wrapping the menu body in `<ng-template matMenuContent>` enables lazy rendering so that the destination buttons are only rendered when the menu is actively opened.

```suggestion
                                <mat-menu #moveMenu="matMenu">
                                  <ng-template matMenuContent>
                                    <button
                                      mat-menu-item
                                      (click)="
                                        movePreparedCandidate($event, {
                                          kind: 'new-after-source',
                                          id: config.sceneIdCounter().toString(),
                                        })
                                      "
                                    >
                                      Move to new scene
                                    </button>
                                    @for (
                                      destination of config.projectConfig.value().storyboard;
                                      track destination.id
                                    ) {
                                      @if (
                                        destination.id !== scene.id &&
                                        config.isGeneratedScene(destination)
                                      ) {
                                        <button
                                          mat-menu-item
                                          [disabled]="
                                            areCandidatesGenerating(destination.id) ||
                                            !!destination.pendingGeneration
                                          "
                                          [matTooltip]="
                                            areCandidatesGenerating(destination.id) ||
                                            destination.pendingGeneration
                                              ? 'Cannot move while this scene is generating'
                                              : ''
                                          "
                                          (click)="
                                            movePreparedCandidate($event, {
                                              kind: 'existing',
                                              sceneId: destination.id,
                                            })
                                          "
                                        >
                                          Move to {{ destination.name }}
                                        </button>
                                      }
                                    }
                                  </ng-template>
                                </mat-menu>
```

victor-paunescu (changes requested):

ui/src/app/storyboard/move-candidate.ts
christophervoelpel: Fixed in the latest commit. When a candidate moves into an existing scene, we now clear that scene's `generationError` and `generationErrorAcknowledged` before adding the moved candidate, so the destination no longer shows a stale "Generation failed" banner or filmstrip badge after the move. Added a test that seeds a destination scene with both fields set and checks they're gone after the move.

ui/src/app/services/remix-engine/remix-engine.ts
christophervoelpel: Fixed. We no longer default `selectedCandidateIndex` to 0 when the candidate list ends up empty (e.g. an empty resumed completion): both `candidates` and `selectedCandidateIndex` are removed from the scene in that case, so there's no out-of-bounds index left for scene-combining to trip over. When the list is non-empty we now also validate any existing selected index is actually in range before keeping it, resetting to 0 only if it's invalid. Added tests for both the empty-completion case and an in-range vs out-of-range selection case.

ui/src/app/storyboard/storyboard.ts
christophervoelpel: Fixed. The tooltip now keeps the "edit of run N" note even when the candidate is back in its original scene, instead of dropping it and showing only "(originally 2A)". Also switched to lowercase "originally" to match the wording used elsewhere. Added a test covering the same-scene case with an edit history.

ui/src/app/storyboard/storyboard.scss
christophervoelpel: Fixed with a small CSS rule: when the edit button isn't present, the move button shifts over into the space it would have used, instead of leaving a gap.

ui/src/app/storyboard/storyboard.ts
christophervoelpel: Fixed. A rejected move (busy source/destination, or an invalid source/destination, e.g. the candidate changed underneath the move, or the target scene no longer exists) now shows a snackbar telling the user the move didn't happen, instead of failing with no feedback. Verified with a parametrized test covering all three rejection reasons, and confirmed by mutation testing that the test actually catches a regression here (silencing the snackbar call made the test fail as expected).

ui/src/app/storyboard/storyboard.html
christophervoelpel: Fixed. The per-destination buttons in the move menu are now wrapped in `<ng-template matMenuContent>`, so they're only built when the menu is actually opened instead of for every candidate in every scene up front.

victor-paunescu (dismissed):

victor-paunescu (approved):

PR #182, opened 2026-09-10
Add bounded prompt dictation and responsive editing

christophervoelpel:
## Summary

Add bounded microphone dictation to Setup, Storyboard and Edit candidate text fields. Dictation is included and enabled by default in deployment and the backend; operators can opt out with `DICTATION_ENABLED=0`. Existing explicit `0` settings stay disabled on redeploy. It uses the fixed Gemini Transcribe adapter with 4 MiB/120-second audio limits and no durable recording storage.

Users review the inserted text before submitting; dictation does not automatically generate or submit anything. Dictation inserts at the caret/selection captured when recording starts, supports Undo, and keeps scene-prompt ownership with the scene when candidate selection changes. The redundant dismiss-message X is removed.

`DICTATION_MODE=SMART` is the default admin configuration: it removes fillers, formats text and applies spoken self-corrections. Operators can set `DICTATION_MODE=VERBATIM` in `config.txt` before deployment for literal transcription. Only these exact uppercase values are accepted; an omitted value defaults to SMART, while empty or invalid values are rejected before provider client construction. The mode is passed to the app only; the model, SDK, request endpoint and microphone UI are unchanged. Redeployment uses the configured value, replacing any prior Cloud Run value. Root/UI READMEs and DEVELOPING document this behavior.

Related editing fixes keep the dictation controls clear of text-area scrollbars, make Storyboard controls reachable at tablet widths, and show exactly one placeholder for an edit regardless of the normal generation candidate count. Narrow screens stack the prompt below the trim inputs and use a single-column product setup. The edit field label is “Describe your edit”.

Trim follow-up: paused/scrubbed main video previews disable native looping so the right trim handle can stay at the endpoint without flashing back to the first frame. Normal playback and thumbnail hover looping are preserved. Full-span trims no longer show a redundant trimmed-duration chip. The measured media duration is not rounded down to the model's nominal duration.

This PR also changes deployment-time GCS bucket CORS configuration so signed-media fetches work from the app's actual Cloud Run service URLs. `deploy/render_cors_origins.py` validates the returned URL annotation and renders its explicit HTTPS origins into `gcs-cors-config.template.json`; `deploy.sh` uses that result when updating the bucket. Missing or malformed metadata fails deployment rather than inventing an origin or using a wildcard. This is a bucket-policy/deployment change, not part of audio transcription itself, and is included here with the deployment fixes exercised during feature testing.

The matching developer-guide wording now uses the renamed “Generate Audio” control. The root and UI READMEs and developer guide explain default enablement, explicit opt-out, user-initiated recording, cloud processing and model usage costs. No recording starts merely because the feature is enabled.

## Victor review follow-up

Provider policy blocks are recognized before candidate-shape validation, including the SDK's `candidates=None` response, and blocked content is never inserted. Missing MediaRecorder support stops acquired tracks and shows a useful error. Stopping a recorder clears its timer before delayed callbacks; size overflow remains a visible recovery error. Corrected the deployment test's shell escaping.

User-feedback follow-up: a normal-flow footer reserves space below the scrollable text, with the mic at the bottom right and Undo/status alongside it. Long error/recovery details open only on request in a bounded dialog. Dictation uses the focused field's caret or selection, including keyboard transfer to its mic; otherwise it appends at the end. No overlay covers text and dictation-disabled layout is preserved.

## Validation

Chrome at 375×812, 768×1024, 1024×768 and 1440×900 verifies all six dictation fields: reserved bottom-right footer, no text/scrollbar overlap, caret insertion and Undo. Repeated right-edge trim dragging stays at the endpoint without looping; full-span trim chips stay hidden. Recorders, transcription and seekable media are synthetic; no real microphone or provider call was made.

## Stack

Stack: depends on #181. Review this PR against its stacked base. After its parent is squash-merged, rebase only this PR's change onto `main` before retargeting. Current GitHub workflows run for PRs targeting `main`, so full CI on this stacked PR follows that retargeting.

The client now automatically stops a two-minute recording at 119 seconds, leaving foreground scheduling/encoding headroom below the unchanged hard 120-second server limit. Shorter 30-second fields are unchanged. The rendered control/service regressions cover delayed stop/final chunks and the hard decoder limit remains enforced.

Real Chrome, synthetic WebAudio through MediaRecorder and the production normalizer: a deliberate 500 ms foreground stall near the deadline produced a 119.400000-second clip that was accepted. This uses a local harness shell, not a personal microphone, paid provider or deployed IAP flow. It does not promise acceptance after arbitrarily delayed background-tab timers. Focused dictation regressions: 26 passed; independent challenge found no surviving code defect.

Release dependency: the CORS correction in this PR is also required for #180 original downloads on app origins not already allowed by the bucket. Include it in the minimum deployable release and verify downloads from each supported origin at deployment. No new deployment/origin smoke was performed here.

Descendants have been rebased to inherit this fix. Full application CI remains a fresh-head requirement after retargeting to `main`; local validation is not an upstream CI claim.

PR head: `f8a7ecb`. Fresh integrated stack at #192 `f454c25`: **645 UI tests in 42 files**, compile, lint and spec typecheck pass on Node22.22.3/npm10.9.3. Production build passes on the identical runtime tree before the final test-only addition. The final backend tree passes **678 tests plus15 subtests** on Python3.13; action signatures, status-viewer tests, ShellCheck, Bash syntax and whitespace checks pass. The separate Python3.11/3.12 CI matrix was not run locally.


miguelfc (approved):

transcription.py
```diff
@@ -0,0 +1,470 @@
...
+  if content is None:
+    if _prompt_block_reason(response) is not None:
+      raise TranscriptionError(
+          'provider_blocked', 502, 'Transcription was blocked'
+      )
+    if _is_completed_empty(response, candidates[0]):
+      return []
+    raise TranscriptionError(
+        'malformed_provider_response', 502, 'Transcription response was invalid'
+    )
```
victor-paunescu, line 401: When Vertex AI blocks an input prompt due to safety filters, `response.candidates` is typically empty (`[]`) while `response.prompt_feedback.block_reason` is set. Currently, line 376 returns `[]` immediately when `candidates` is empty, causing `parse_response()` to return an empty string and the endpoint to return HTTP 200 `{'text': ''}`.

Prompt-level and candidate-level safety or policy blocks (`SAFETY`, `BLOCKLIST`, `PROHIBITED_CONTENT`, `SPII`) should be mapped directly to `provider_blocked` (HTTP 502) rather than silently returning an empty transcript or raising `malformed_provider_response`.

```suggestion
  if not candidates:
    if _prompt_block_reason(response) is not None:
      raise TranscriptionError(
          'provider_blocked', 502, 'Transcription was blocked'
      )
    return []
  candidate = candidates[0]
  candidate_finish_reason = _enum_value(getattr(candidate, 'finish_reason', None))
  if (
      _prompt_block_reason(response) is not None
      or candidate_finish_reason in (
          'SAFETY', 'BLOCKLIST', 'PROHIBITED_CONTENT', 'SPII'
      )
  ):
    raise TranscriptionError(
        'provider_blocked', 502, 'Transcription was blocked'
    )
  content = getattr(candidate, 'content', None)
  if content is None:
    if _is_completed_empty(response, candidate):
      return []
    raise TranscriptionError(
        'malformed_provider_response', 502, 'Transcription response was invalid'
    )
```

ui/src/app/shared/dictation/dictation.ts
```diff
@@ -0,0 +1,426 @@
...
+  dismiss(owner: string): void {
+    if (this.activeOwner === owner) this.cancel(owner);
+    else if (this.state().owner === owner)
+      this.setState({status: 'idle', elapsedSeconds: 0});
+  }
+
+  private supportedMimeType(): string | undefined {
+    if (typeof MediaRecorder.isTypeSupported !== 'function') return undefined;
+    return this.mimeTypes.find(type => MediaRecorder.isTypeSupported(type));
+  }
```
victor-paunescu, line 247: Evaluating property access on an undeclared global identifier raises a runtime `ReferenceError` (`ReferenceError: MediaRecorder is not defined`) in non-supporting browsers or test environments. Checking `typeof MediaRecorder === 'undefined'` first ensures safe fallback.

```suggestion
  private supportedMimeType(): string | undefined {
    if (
      typeof MediaRecorder === 'undefined' ||
      typeof MediaRecorder.isTypeSupported !== 'function'
    ) {
      return undefined;
    }
    return this.mimeTypes.find(type => MediaRecorder.isTypeSupported(type));
  }
```

ui/src/app/shared/dictation/dictation.ts
```diff
@@ -0,0 +1,426 @@
...
+      return;
+    }
+    if (this.recorder.state === 'recording') {
+      this.setState({
+        status: 'transcribing',
+        owner,
+        elapsedSeconds: this.elapsedSeconds(),
+      });
+      this.recorder.stop();
+    }
```
victor-paunescu, line 206: `this.timer` is only cleared when `finishRecording()` is invoked asynchronously in the `onstop` callback. If the 250ms interval fires during this asynchronous gap, `updateTimer` executes and resets `status` back to `'recording'`, clobbering `'transcribing'` in the UI. Clearing `this.clearTimer()` immediately on stop eliminates this race condition.

```suggestion
    if (this.recorder.state === 'recording') {
      this.clearTimer();
      this.setState({
        status: 'transcribing',
        owner,
        elapsedSeconds: this.elapsedSeconds(),
      });
      this.recorder.stop();
    }
```

ui/src/app/shared/dictation/dictation-control.ts
```diff
@@ -0,0 +1,492 @@
...
+    effect(() => {
+      const current = this.state();
+      if (current.eventId === this.handledEventId || current.status !== 'error')
+        return;
+      this.handledEventId = current.eventId;
+      this.recovery.set({
+        kind: 'failed',
+        message: current.error ?? 'Transcription failed. Please try again.',
+      });
+    });
```
victor-paunescu, line 293: The recovery effect checks `current.status !== 'error'`. When `DictationService` emits `status: 'overflow'` because the audio recording exceeded the 4 MiB payload cap, this effect returns early. As a result, the microphone turns off and returns to the idle state without presenting any error or recovery message to the user.

```suggestion
    effect(() => {
      const current = this.state();
      if (
        current.eventId === this.handledEventId ||
        (current.status !== 'error' && current.status !== 'overflow')
      )
        return;
      this.handledEventId = current.eventId;
      this.recovery.set({
        kind: 'failed',
        message:
          current.error ??
          (current.status === 'overflow'
            ? 'Recording exceeded the 4 MiB audio limit.'
            : 'Transcription failed. Please try again.'),
      });
    });
```

test/test_deploy_safety.py
```diff
@@ -198,3 +322,92 @@ def test_no_infinite_gunicorn_timeout():
...
+  """Execute the deploy.sh mode block without running the deploy."""
+  text = _deploy_sh()
+  match = re.search(
+      r'(?ms)^(DICTATION_MODE="\$\{DICTATION_MODE-SMART\}"\n'
+      r'if \[\[.*?^fi)$',
+      text,
+  )
+  assert match
+  block = match.group(1) + '\nprintf "%s\n" "$DICTATION_MODE"\n'
+  environment = os.environ.copy()
```
victor-paunescu, line 401: In Python string literals, `"%s\n"` interpolates an actual ASCII newline character (0x0A) rather than escaping `\n` for the shell. Line 365 in the sibling test correctly uses `"%s\\n"` so `printf` receives the intended format string.

```suggestion
  assert match
  block = match.group(1) + '\nprintf "%s\\n" "$DICTATION_MODE"\n'
  environment = os.environ.copy()
```

victor-paunescu (changes requested):

transcription.py
christophervoelpel: Fixed. `_response_parts` now checks `_prompt_block_reason` first and `_is_blocked_finish_reason` on the first candidate before falling through to the empty-completion path, so SAFETY/BLOCKLIST/PROHIBITED_CONTENT/SPII all raise `provider_blocked` (502) instead of returning an empty transcript. Covered by `test_prompt_feedback_block_with_no_candidates_is_safe_502`, `test_prompt_feedback_block_with_missing_candidates_is_safe_502`, `test_prompt_feedback_block_wins_over_stop_content`, and `test_blocked_finish_reason_is_safe_even_with_content` (parametrized over all four reasons, with and without content).

ui/src/app/shared/dictation/dictation.ts
christophervoelpel: Fixed in `315b1f2` (dictation.ts, `supportedMimeType()`): it now checks `typeof MediaRecorder === 'undefined'` before calling `MediaRecorder.isTypeSupported`, so environments without `MediaRecorder` no longer throw.

ui/src/app/shared/dictation/dictation.ts
christophervoelpel: Fixed in `315b1f2`: `stop()`, the overflow path, and the failure path in dictation.ts now all call `clearTimer()` before `setState(...)`, so the 250ms timer can no longer fire during the gap and clobber the status back to "recording".

ui/src/app/shared/dictation/dictation-control.ts
christophervoelpel: Fixed in `315b1f2`: the state-watching effect in dictation-control.ts now also matches `status === 'overflow'` and sets the same recovery/failed messaging path used for errors, so hitting the audio cap shows the same review/dismiss UI instead of silently going idle.

test/test_deploy_safety.py
christophervoelpel: Fixed. Both new deploy-validation tests (`test_dictation_deploy_validation_executes_default_and_opt_out` and `test_dictation_mode_deploy_validation_executes_exact_allowlist`) now use `'\nprintf \"%s\\n\" ...'` with the escaped `\n`, consistent with the existing pattern.

victor-paunescu (dismissed):

victor-paunescu (dismissed):

victor-paunescu (approved):

PR #184, opened 2026-09-10
Document contributor testing and documentation workflows

christophervoelpel:
## Summary

Add a practical contributor workflow for UI testing, feature documentation and release-note preparation.

- `docs/testing.md`: fresh-checkout commands, phone/tablet/desktop viewports, real interaction checks, cloud/provider safety and a reproducible evidence template.
- `docs/documentation.md`: update the existing walkthrough in place, use real screenshots when useful, and keep a canonical changelog separate from the README teaser.
- Three small reusable skills, linked from `DEVELOPING.md`, share the guides instead of duplicating them.

`docs/skills/*/SKILL.md` introduces a provider-neutral, opt-in convention for agent-readable contributor instructions; these are plain Markdown guides, not executable hooks, and do not require a particular assistant or installation.

This adds instructions, not a browser-testing framework, deployment automation or a versioned release.

## Victor review follow-up

Walkthrough media examples now resolve relative to their containing document. Feature notes go under `Unreleased`, not a fabricated release. The testing guide links to the reusable UI-testing skill without duplicating its instructions.

## Validation

Node 22.22.3/npm 10.9.3. The final #180–#184 source tree (combined head `a0b1753`, byte-identical verification tree `ce14111`) passes all 507 UI tests in 34 files, compile, spec typecheck, lint and production build. Production dependency audit reports zero vulnerabilities.

The combined stack also passes 644 backend tests plus 15 subtests, action-signature validation, Bash syntax and ShellCheck. Existing SDK deprecation/signature warnings remain non-failing.

Independent documentation review and Markdown/frontmatter/local-link checks passed. This PR remains six Markdown files only. It adds no test runner, deployment or release.

No live Firestore changes, provider generation, deployment or reviewer messages were made for this update.

## Stack

Stack: depends on #183. Review this PR against its stacked base. After its parent is squash-merged, rebase only this PR's change onto `main` before retargeting. Current GitHub workflows run for PRs targeting `main`, so full CI on this stacked PR follows that retargeting.

## Review-stack update — 13 September

Rebased only to inherit the review fixes in ancestor PRs. A range-diff confirms this PR's own feature commits are unchanged; no additional finding or feature work was added here. Fresh combined-stack validation and exact head are recorded in the publication follow-up below. Application CI still needs to run after the parent merges and this PR is retargeted to `main`.

## Current publication validation

[...]

PR #185, opened 2026-09-11
Cache candidate videos locally between scene visits

christophervoelpel:
## Summary

Keep selected or hovered candidate videos in a bounded local browser cache so switching scenes does not repeatedly download the same clips. This is a separate feature above #184; it does not change original downloads, composition rendering or candidate audio intent.

- Native Cache Storage, no service worker or new dependencies. Complete HTTP 200 responses only; partial responses never become cached videos.
- Fixed seven-day expiry. Best-effort limits: 64 MiB per video, 256 MiB total, 64 entries; evict expired/oldest entries first.
- Stable origin/bucket/project/object-path keys rather than expiring signed URLs. Only selected/hovered videos download; unselected thumbnail videos do not eagerly load.
- Initial full download shows a loading indicator; later visits and page reloads play local blob URLs. Cache failures, unsupported storage and oversized files fall back to normal streaming.
- Archive and successful project deletion invalidate matching local entries, fencing late in-flight writes within the current page. Archived previews do not repopulate the cache.
- Shared downloads and reference-counted object URLs avoid duplicate fetches and premature revocation. Scene edits and signed-URL refreshes do not reset playback; changing/destroying a consumer stops and releases old media.

The cache is opportunistic, not offline project access or a security-revocation mechanism. Other devices cannot be purged by a local action, and the browser can evict entries earlier. README documents this boundary.

## Validation

At `467ef1d`, Node 22.22.3/npm 10.9.3: 531 UI tests in 36 files, compile, lint, spec typecheck and production build pass. The inherited backend tree passes 644 tests plus 15 subtests, action signatures, Bash syntax and ShellCheck. No production dependency changes.

Three Luna agents implemented/reviewed the cache work, with root review and real Chrome verification. Regressions cover shared leases, stable keys, expiry, limits/fallbacks, invalidation during fetch/write, stale completions, archive/delete integration and directive lifetime. Root review caught and fixed premature shared-URL revocation and playback resets on unrelated config changes.

Real Chrome at 375×812, 768×1024, 1024×768 and 1440×900: A → B → A → reload produces exactly two video requests total, with no eager unselected downloads. Blob playback and seeking work, and editing the prompt preserves the URL, playback position and load count. Synthetic media and a local stub API only; no provider, microphone, Firestore or deployment was used.

The final cached-player build also passes the combined four-width trim and six-field dictation browser checks, including caret insertion, Undo, reserved footer bounds and held end-handle dragging.

## Stack

Depends on #184. Review against `publish/contributor-guides-20260910`. After the parent is squash-merged, rebase only this feature onto `main` before retargeting; the repository's full CI runs on PRs targeting `main`.

Documentation now precisely states that in-flight invalidation fencing is page/service-instance-local. Another open tab can finish a pending fetch and repopulate shared Cache Storage; other browsers/offline devices can retain bytes until eviction. No runtime change or cross-tab purge mechanism was added. The inherited dictation headroom fix is carried by the rebase.

PR head: `14ad070`. Fresh integrated stack at #192 `f454c25`: **645 UI tests in 42 files**, compile, lint and spec typecheck pass on Node22.22.3/npm10.9.3. Production build passes on the identical runtime tree before the final test-only addition. The final backend tree passes **678 tests plus15 subtests** on Python3.13; action signatures, status-viewer tests, ShellCheck, Bash syntax and whitespace checks pass. The separate Python3.11/3.12 CI matrix was not run locally.


ui/src/app/shared/candidate-video/candidate-video.directive.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+    this.lease = lease;
+    this.element.nativeElement.src = lease.url;
+    this.loadVideo();
+    this.loading.set(false);
+    if (this.candidateVideoHover) {
+      const video = this.element.nativeElement;
+      video.muted = true;
+      video.playbackRate = 2;
+      try {
```
victor-paunescu: `acquire()` currently lacks a `try ... finally` block around acquisition. If `this.cache.acquire(...)` rejects (for example on a network failure, abort, or storage error), `this.loading.set(false)` is never reached, leaving `candidateVideo.loading()` permanently `true` and locking the `<mat-spinner>` over the player.

Additionally, assigning `this.lease = lease;` without releasing any pre-existing lease leaks the reference count in `CandidateVideoCacheService`, preventing `URL.revokeObjectURL()` from ever being invoked when elements are re-acquired.

Wrap acquisition in `try ... finally`, release any prior lease before assignment, and ensure empty URLs clear `src`:

```suggestion
    const requestId = ++this.requestId;
    this.loading.set(true);
    try {
      const lease = await this.cache.acquire(
        this.candidateVideoCacheScope ?? {bucket: '', projectId: ''},
        media,
        this.candidateVideoPersist && !!this.candidateVideoCacheScope,
      );
      if (requestId !== this.requestId) {
        lease.release();
        return;
      }
      this.lease?.release();
      this.lease = lease;
      if (lease.url) {
        this.element.nativeElement.src = lease.url;
      } else {
        this.element.nativeElement.removeAttribute('src');
      }
      this.loadVideo();
      if (this.candidateVideoHover) {
        const video = this.element.nativeElement;
        video.muted = true;
        video.playbackRate = 2;
        try {
          const playback = video.play();
          if (playback) void playback.catch(() => undefined);
        } catch {
          // A preview may be unavailable in a browser without media playback.
        }
      }
    } catch {
      // Acquisition failures fall back to unbuffered/native media handling.
    } finally {
      if (requestId === this.requestId) {
        this.loading.set(false);
      }
    }
```

ui/src/app/shared/candidate-video/candidate-video.directive.ts
```diff
@@ -0,0 +1,134 @@
...
+    this.lease?.release();
+    this.lease = undefined;
+    this.loading.set(false);
+    if (clearElement) {
+      const video = this.element.nativeElement;
+      video.pause();
+      video.removeAttribute('src');
+      this.loadVideo();
+    }
+  }
```
victor-paunescu, line 144: In `release()`, `this.lease?.release()` is currently invoked before lines 119–124 clear the element. When the lease reference count reaches zero, `CandidateVideoCacheService.release()` synchronously calls `URL.revokeObjectURL(entry.url)`. Because `video.src` is still pointing to that blob URL when it is revoked, WebKit/Safari fires an asynchronous `MEDIA_ERR_SRC_NOT_SUPPORTED` error event on the media element.

Clearing the element's `src` attribute before releasing the lease ensures the object URL is detached before revocation:

```suggestion
  private release(clearElement = false): void {
    this.requestId++;
    if (clearElement) {
      const video = this.element.nativeElement;
      video.pause();
      video.removeAttribute('src');
      this.loadVideo();
    }
    this.lease?.release();
    this.lease = undefined;
    this.loading.set(false);
  }
```

ui/src/app/services/media/candidate-video-cache.ts
```diff
@@ -0,0 +1,446 @@
...
+              const blob = await this.readResponse(cached);
+              if (this.version(scope.projectId, file.path!) === version) {
+                return {blob, fallbackUrl: ''};
+              }
+            } catch {
+              // A corrupt or unexpectedly large entry is discarded below.
+            }
+          }
+          await cache.delete(request);
+        }
```
victor-paunescu, line 174: When reading a valid cached response, if invalidation occurs concurrently so that `this.version(...) === version` evaluates to `false`, execution exits the `try` block without returning and proceeds to line 176. It then performs a full HTTP network fetch and stream read of up to 64MB, only to check the version again at line 184 and discard the downloaded blob.

If the version changed during the cache read, this acquisition has already been invalidated. Return the fallback URL immediately to avoid an expensive and redundant network download:

```suggestion
            try {
              const blob = await this.readResponse(cached);
              if (this.version(scope.projectId, file.path!) === version) {
                return {blob, fallbackUrl: ''};
              }
              return {fallbackUrl: await this.resolve(file)};
            } catch {
              // A corrupt or unexpectedly large entry is discarded below.
            }
          }
          await cache.delete(request);
```

ui/src/app/services/media/candidate-video-cache.ts
```diff
@@ -0,0 +1,446 @@
...
+  private mutationQueue: Promise<void> = Promise.resolve();
+
+  async acquire(
+    scope: CandidateCacheScope,
+    file: MediaRef | null | undefined,
+    persist: boolean,
+  ): Promise<CandidateVideoLease> {
+    if (!persist || !file?.path) {
+      return this.directLease(await this.resolve(file));
+    }
```
victor-paunescu: `acquire()` validates `persist` and `file?.path`, but does not verify that `scope.bucket` and `scope.projectId` are non-empty. If an empty scope is passed with `persist = true`, `cacheRequest` constructs a URL with empty path segments (`/__scene_machine_candidate_video_cache__/v1///path`).

When `parseCacheKey()` parses these URLs, `parts.some(part => !part)` fails, returning `undefined`. Consequently, these entries can never be matched or evicted by `deleteMatching()` during candidate or project invalidation.

Validate that both bucket and projectId are present before attempting persistent caching:

```suggestion
    if (
      !persist ||
      !file?.path ||
      !scope?.bucket ||
      !scope?.projectId
    ) {
      return this.directLease(await this.resolve(file));
    }
```

ui/src/app/storyboard/storyboard.scss
```diff
@@ -946,6 +948,14 @@
           object-fit: contain;
         }
 
+        .candidate-video-loading {
+          position: absolute;
+          top: 50%;
+          left: 50%;
+          transform: translate(-50%, -50%);
+          z-index: 3;
+        }
```
victor-paunescu, line 958: Because `.candidate-video-loading` has `z-index: 3`, it renders above `.video-overlay`. Without `pointer-events: none;`, clicks within the center 24px of the video hit the spinner element and bubble past the overlay click handler (`toggleVideoPlay()`), creating an unclickable dead zone.

Add `pointer-events: none;` to pass pointer interactions through to the video overlay:

```suggestion
        .candidate-video-loading {
          position: absolute;
          top: 50%;
          left: 50%;
          transform: translate(-50%, -50%);
          z-index: 3;
          pointer-events: none;
        }
```

ui/src/app/services/media/candidate-video-cache.spec.ts
```diff
@@ -0,0 +1,551 @@
...
+    expect(second.url).toBe('blob:lease');
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    expect(mediaServiceMock.resolve).toHaveBeenCalledTimes(1);
+    expect([...cache.values.keys()][0]).not.toContain('signed.example');
+    second.release();
+  });
+
+  it('discards cache entries with invalid metadata before fetching fresh bytes', async () => {
+    cache.values.set(
+      'http://localhost:3000/__scene_machine_candidate_video_cache__/v1/bucket-a/project-a/videos/a.mp4',
```
victor-paunescu: In the tests `'discards cache entries with invalid metadata before fetching fresh bytes'` and `'does not serve a cache entry dated in the future'`, the mock key in `cache.values` uses unencoded slashes (`videos/a.mp4`). However, `CandidateVideoCacheService.cacheRequest` uses `encodeURIComponent(path)`, producing `videos%2Fa.mp4`.

Because `MemoryCache.match` looks up by exact URL string, `cache.match(request)` returns `undefined` immediately. The tests pass only because a cache miss triggers `fetchMock`, without actually exercising or verifying the invalid metadata or future timestamp expiration branches.

Update the mock keys to use `videos%2Fa.mp4`:

```suggestion
    cache.values.set(
      'http://localhost:3000/__scene_machine_candidate_video_cache__/v1/bucket-a/project-a/videos%2Fa.mp4',
```

victor-paunescu (changes requested):

ui/src/app/shared/candidate-video/candidate-video.directive.ts
christophervoelpel: Fixed in 847a69d. Acquisition now uses try/catch/finally with a request-id guard, so rejection clears only the current request's loading state. Existing leases are detached and released before replacement, and an empty URL removes src. Regression tests cover rejection, stale rejection while a newer request is pending, reacquisition cleanup, and empty URLs, including a real cache-service URL-encoding rejection.

All six review points are addressed in this push. The 32 focused tests and full 542-test UI suite passed locally, along with typechecks, lint and the production build; fresh CI is green. Ready for another review and approval if you are happy with the changes.

ui/src/app/shared/candidate-video/candidate-video.directive.ts
christophervoelpel: Fixed in 847a69d. Both teardown and lease replacement now pause the video, remove src and call load() before releasing the old lease. Tests assert detach-before-release ordering, including real cache-service object-URL revocation. Safari itself was not run, so this verifies the lifecycle ordering rather than claiming a Safari-specific reproduction.

ui/src/app/services/media/candidate-video-cache.ts
christophervoelpel: Fixed in 847a69d. If invalidation changes the version while a valid cached response is being read, acquisition now returns the signed fallback immediately. The regression test uses a paused ReadableStream and verifies no bulk fetch, object URL or stale cache entry. Removing the early return makes that test fail on an unwanted fetch.

ui/src/app/services/media/candidate-video-cache.ts
christophervoelpel: Fixed in 847a69d. Persistent caching now requires both a non-empty bucket and projectId; otherwise acquisition returns the signed URL directly. The regression test covers each missing scope field and verifies that Cache Storage is not opened and no video fetch or cache entry is created. Removing the guard makes the test fail.

ui/src/app/storyboard/storyboard.scss
christophervoelpel: Fixed in 847a69d with pointer-events: none on the loading spinner. In Chrome, with a deliberately delayed synthetic media response, hit-testing at the visible spinner's centre reached the underlying video-overlay. The spinner disappeared after the cached video source attached.

ui/src/app/services/media/candidate-video-cache.spec.ts
christophervoelpel: Fixed in 847a69d. Both fixtures now use the actual test origin and encoded videos%2Fa.mp4 key. The tests inspect the matched response's invalid/future timestamp after acquisition and assert deletion plus a fresh fetch, so an accidental cache miss can no longer satisfy them.

victor-paunescu (approved):

PR #187, opened 2026-09-12
Cache homepage and storyboard thumbnails locally

christophervoelpel:

This PR's code is unchanged at `3e1586f`. Fresh integrated validation at #192 `dbe6f74`: **653 UI tests in 42 files**, compile, lint, spec typecheck, status-viewer tests and production build passed on Node 22.22.3/npm 10.9.3. The identical final backend tree passed **678 tests** on Python 3.13; action signatures, ShellCheck, Bash syntax and diff checks passed. The sandboxed build aborted; the same production build passed outside the sandbox.

The original bounded signing-recovery fix remains here. The separate initial cache-scope flicker fix and strengthened timer-cleanup assertion live in #189 with the performance work; descendants inherit them. No new cache budget or eviction behavior was added.

## Summary

Cache homepage project thumbnails, storyboard filmstrip images and candidate
preview images locally. Keep the image cache separate from candidate videos:
**1 MiB per image, 64 MiB total, at most 1,024 entries, seven-day expiry**.

- Reuse the video cache machinery through a shared engine; preserve the existing
  video cache namespace and public service API. Raise the total video budget to
  **512 MiB**, keeping **64 MiB per video**, **64 entries** and seven-day expiry.
- Key images by bucket, project and immutable object path. Candidate selection
  switches to the correct thumbnail; late responses cannot replace a newer image.
- Load images within 200 px of the viewport, release object URLs when no longer
  displayed, and serialize cache writes across tabs where Web Locks is available.
- Invalidate matching images on candidate archive and successful project delete.
  Archived previews display without repopulating persistent cache.
- Remove the homepage's duplicate reference-image load when a thumbnail exists.
  Keep the homepage image-only, including on hover: show the existing static
  placeholder when no thumbnail or reference image exists. Keep image URL
  signing batched. Storyboard video previews are unchanged.
- Document cache limits, fallback behavior and immutable-path requirements.

This deliberately does not change project APIs, resize stored assets, introduce
offline project access, or rewrite application bundles. Full-size editing
reference images are outside thumbnail caching. Oversized images and unavailable
browser storage fall back to ordinary URLs.

## Validation

Node **22.22.3**, npm **10.9.3**:

- 562 UI tests across 38 files, lint and spec typecheck passed.
- Production build passed.
- 651 backend tests plus 15 subtests passed; status-viewer check passed.
- Video cache regressions cover the updated 512 MiB total budget. Added service,
  directive, homepage and storyboard regressions for size/TTL limits, concurrent writes,
  invalidation races, candidate changes and completed-generation thumbnails.
- Provided-video and generated video-only homepage cards render a static
  placeholder with no video element, including on hover.
- Independent skeptical review: no reproducible application defects remain.

Real Chrome **152.0.7977.83**, built Angular UI, local stub API and repository
sample images (no provider, microphone or cloud data):

| 24-project homepage at 1440 × 900 | Before | After |
| --- | ---: | ---: |
| Initial image requests | 46 | 6 |
| Initial image bytes | 5,745,032 | 749,352 |
| Repeat-visit image requests | 46 | 0 |
| Repeat-visit image bytes | 5,745,032 | 0 |
| Initial fallback-video requests | 1 | 0 |

The fixture serves image responses with `Cache-Control: no-store` to isolate the
application cache from ordinary HTTP caching. These are image-transfer figures
for this controlled sample, not a production load-time claim.

Chrome also checked candidate A → B → A, a delayed A response after selecting B,
storyboard reload, archive → save → reload → homepage, project deletion,
image-only homepage hover with no video element or video request, and phone/tablet/desktop
layouts at 375, 768, 1024 and 1440 px (100% zoom).

Chrome evidence is from the image-only homepage follow-up. The later video
budget-only change was verified with the full UI suite and dedicated below,
exactly-at and above-512-MiB boundary tests; no additional browser/provider run.

## Stack / merge

Depends on #186 (`test/omni-audio-intent-matrix-20260911`). Kept as a separate feature; the 13 September review fixes were propagated through its ancestors. After the parent merges,
rebase this PR's own commit onto `main`, retarget, and run fresh upstream CI.

Browser storage is best-effort. Invalidation occurs when the current page observes
archive/delete; other browsers or offline devices can retain their bytes until
expiry/eviction. This is not a remote security purge. No deployment is included.

Fixed same-media thumbnail recovery after rejected or empty cache/signing acquisition. Recovery is bounded (250 ms, 1 s, 4 s delays), not retried on each change-detection pass. Media changes and destruction cancel pending recovery; stale/empty leases are released. After exhaustion, a longer outage may require remount/reload.

Rendered Angular directive tests exercise unchanged-path recovery, bounded exhaustion, switching and destruction; 11 focused tests passed. The cache boundary is mocked in these tests, so this is not new deployed-browser outage evidence. Independent source challenge found no surviving defect. Descendants inherit the same fix; it is not duplicated in #191.


christophervoelpel:
## Live Before / After cache benchmark — 12 September 2026

Adding the paired cloud results here because this PR contains the shared
video/image cache engine. **This measures the complete stack, not #187 in
isolation:** candidate-video caching in #185, thumbnail/shared caching in
#187, and delivery/project-loading improvements in #189 are included.

- Before: upstream main at `844b04766729704a3d2f38318caab3a3c28dd744`.
- After: the stack through #190 at `f5a2a73e70213e49d09e9b2d8e2352ed6c033509`.
- Two fresh GCP deployments with matching configuration and media bytes.
  Homepage: ten projects with ten distinct 720p candidate clips. Storyboard:
  a separate five-scene/two-candidate-per-scene fixture, added after the
  homepage measurements.
- Same authenticated Chrome profile, initial 1440×900 CSS viewport, 100% page
  zoom, no CPU/network throttling. Backend cold starts were not measured.

### Repeat-visit medians

| Page | Runs per side | Transferred Before → After | Last API/media request end Before → After |
| --- | ---: | ---: | ---: |
| Homepage | 5 | 267 KB → 9.5 KB | 2.94 s → 0.72 s |
| Storyboard | 3 | 5.25 MB → 9.2 KB | 3.26 s → 0.68 s |

Decimal KB/MB; transferred bytes come from Chrome HAR `_transferSize`.
The timing column is a network-request endpoint, **not an application-ready
or interactivity measurement**. Before's partial video downloads include
browser cancellations; their actual transferred bytes and request-end times
remain included. No HTTP-error/status-0 responses occurred.

The warm transfer ranges were 266,517–266,652 B Before versus 9,532–9,538 B
After for Homepage, and 5,164,862–5,263,068 B Before versus 9,183–15,874 B
After for Storyboard. After's higher Storyboard sample included two signing
requests and was retained.

### Scene return and interpretation

Returning from Scene 2 to Scene 1 transferred **1,671,554 B Before versus
51 B After**. After made no video or thumbnail network requests; both sides
retained a 51-byte PATCH. This is a counts/bytes check, excluded from reload
timing aggregates.

Only the initial viewport and visited media were measured. After deferred
three below-fold homepage thumbnails that Before fetched, so this is not a
full-page/all-candidate download comparison. The warm homepage result covers
the seven thumbnails actually loaded. The decoded project-list payloads
matched; smaller API transfers also reflect compression, not a smaller schema.
Storyboard's project-data API median was essentially unchanged (327 ms versus
331 ms); its main observed saving was avoiding repeated media transfers.

The fixtures use **generated candidate records**, not uploaded-video scenes;
these results do not validate persistent caching for uploaded source videos.
Independent review verified payload/media identity, aggregate arithmetic and
the distinction between HTTP failures and browser media cancellations. Raw
HARs and traces remain private because they contain signed media URLs and
session details.

ui/src/app/shared/thumbnail-image/thumbnail-image.directive.ts
```diff
@@ -0,0 +1,192 @@
...
+const THUMBNAIL_RECOVERY_DELAYS_MS = [250, 1000, 4000] as const;
+
+/**
+ * Loads a thumbnail only when its image is near the viewport, while allowing
+ * the cache service to persist the downloaded bytes and share its lease.
+ */
+@Directive({
+  selector: 'img[appThumbnailImage]',
+  standalone: true,
+})
```
victor-paunescu, line 41: Per the project's Angular guidelines (`ui/.agent/rules/angular.md`), `@HostListener` decorators should be avoided in favor of the `host` property in the `@Directive` decorator, and `standalone: true` is redundant as it is the default in Angular v20+.

```suggestion
import {
  Directive,
  ElementRef,
  Input,
  OnChanges,
  OnDestroy,
  inject,
} from '@angular/core';
import {
  CandidateCacheScope,
  CandidateVideoLease,
} from '../../services/media/candidate-video-cache';
import {ThumbnailCacheService} from '../../services/media/thumbnail-cache';
import {MediaRef} from '../../services/media/media';

const THUMBNAIL_RECOVERY_DELAYS_MS = [250, 1000, 4000] as const;

/**
 * Loads a thumbnail only when its image is near the viewport, while allowing
 * the cache service to persist the downloaded bytes and share its lease.
 */
@Directive({
  selector: 'img[appThumbnailImage]',
  host: {
    '(load)': 'onLoad()',
  },
})
```

ui/src/app/shared/thumbnail-image/thumbnail-image.directive.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  ngOnDestroy(): void {
+    this.disposed = true;
+    this.stopObserving();
+    this.cancelRecovery();
+    this.release(true);
+  }
+
+  @HostListener('load')
+  onLoad(): void {
```
victor-paunescu: With the listener declared on `host: { '(load)': 'onLoad()' }`, remove the `@HostListener('load')` decorator from the method declaration.

```suggestion
  onLoad(): void {
```

ui/src/app/shared/thumbnail-image/thumbnail-image.directive.ts
```diff
@@ -0,0 +1,192 @@
...
+    if (
+      !this.lease ||
+      !image.getAttribute('src') ||
+      currentSrc !== this.assignedSrc
+    ) {
+      return;
+    }
+    image.classList.add('loaded');
+    this.thumbnailImageContainer?.classList.add('high-res-loaded');
+  }
```
victor-paunescu, line 96: If an image returns HTTP 200 with corrupt or zero-byte contents, the browser triggers the `load` event with `naturalWidth === 0`. Marking the parent container as `high-res-loaded` unconditionally hides the low-res fallback thumbnail, resulting in a blank display. Verifying `image.naturalWidth > 0` prevents hiding the fallback prematurely.

```suggestion
    const image = this.element.nativeElement;
    const currentSrc = image.currentSrc || image.src;
    if (
      !this.lease ||
      !image.getAttribute('src') ||
      currentSrc !== this.assignedSrc ||
      image.naturalWidth <= 0
    ) {
      return;
    }
    image.classList.add('loaded');
    this.thumbnailImageContainer?.classList.add('high-res-loaded');
  }
```

ui/src/app/shared/thumbnail-image/thumbnail-image.directive.ts
```diff
@@ -0,0 +1,192 @@
...
+      this.recoveryAttempt = 0;
+      this.lease = lease;
+      this.element.nativeElement.decoding = 'async';
+      this.element.nativeElement.src = lease.url;
+      this.assignedSrc = this.element.nativeElement.src;
+    } catch {
+      if (this.disposed || requestId !== this.requestId) return;
+      this.scheduleRecovery(requestId);
+    }
+  }
```
victor-paunescu, line 164: If an unhandled exception occurs after `this.cache.acquire()` resolves (for example during DOM property assignment), `lease` remains allocated in the browser without being released, leaking the associated `Blob` and `URL.createObjectURL`. Declaring `lease` in function scope and releasing it in `catch`, while releasing any existing `this.lease` prior to reassignment, ensures memory is freed deterministically.

```suggestion
  private async acquire(requestId: number): Promise<void> {
    const media = this.media;
    if (!media || this.disposed) return;
    let lease: CandidateVideoLease | undefined;
    try {
      lease = await this.cache.acquire(
        this.thumbnailCacheScope ?? {bucket: '', projectId: ''},
        media,
        this.thumbnailImagePersist && !!this.thumbnailCacheScope,
      );
      if (this.disposed || requestId !== this.requestId) {
        lease?.release();
        return;
      }
      if (!lease?.url) {
        lease?.release();
        this.scheduleRecovery(requestId);
        return;
      }
      this.recoveryAttempt = 0;
      this.lease?.release();
      this.lease = lease;
      this.element.nativeElement.decoding = 'async';
      this.element.nativeElement.src = lease.url;
      this.assignedSrc = this.element.nativeElement.src;
    } catch {
      lease?.release();
      if (this.disposed || requestId !== this.requestId) return;
      this.scheduleRecovery(requestId);
    }
  }
```

ui/src/app/services/media/media-cache-engine.ts
```diff
@@ -0,0 +1,470 @@
...
+      const metadata = response && this.metadata(response);
+      if (!metadata || !this.isFresh(metadata.downloadedAt)) {
+        await cache.delete(request);
+        continue;
+      }
+      entries.push({request, metadata});
+    }
+    entries.sort((a, b) => a.metadata.downloadedAt - b.metadata.downloadedAt);
+    return entries;
+  }
```
victor-paunescu, line 308: `validEntries()` executes sequentially inside the exclusive `${cacheName}:mutations` Web Lock. For `ThumbnailCacheService` with up to 1,024 entries, sequentially awaiting `cache.match(request)` incurs over 1,000 asynchronous IPC round-trips to the browser storage process, blocking all concurrent cache operations across tabs. Querying responses in parallel via `Promise.all` significantly reduces lock contention.

```suggestion
  private async validEntries(cache: Cache): Promise<CacheEntry[]> {
    const requests = await cache.keys();
    const responses = await Promise.all(
      requests.map(request => cache.match(request)),
    );
    const entries: CacheEntry[] = [];
    for (let i = 0; i < requests.length; i++) {
      const request = requests[i];
      const response = responses[i];
      const metadata = response && this.metadata(response);
      if (!metadata || !this.isFresh(metadata.downloadedAt)) {
        await cache.delete(request);
        continue;
      }
      entries.push({request, metadata});
    }
    entries.sort((a, b) => a.metadata.downloadedAt - b.metadata.downloadedAt);
    return entries;
  }
```

ui/src/app/homepage/homepage.ts
```diff
@@ -166,12 +162,22 @@ export class Homepage {
 
     return {
       ...thumb,
-      showReference: thumb.referenceImage !== undefined,
-      showVideo: !hasThumb && !thumb.referenceImage && !!thumb.videoUrl,
-      showPlaceholder: !hasThumb && !thumb.referenceImage && !thumb.videoUrl,
+      showReference: !hasThumb && thumb.referenceImage !== undefined,
+      showPlaceholder: !hasThumb && !thumb.referenceImage,
     };
```
victor-paunescu, line 170: Checking `thumb.referenceImage !== undefined` causes empty objects (`{}`) or records with undefined `path` and `url` to evaluate to `showReference: true`, mounting an `` tag without a valid source instead of displaying the placeholder. Matching the check in `storyboard.ts` (`!!(referenceImage?.path || referenceImage?.url)`) provides consistent fallback behavior.

```suggestion
    return {
      ...thumb,
      showReference:
        !hasThumb &&
        !!(thumb.referenceImage?.path || thumb.referenceImage?.url),
      showPlaceholder:
        !hasThumb &&
        !(thumb.referenceImage?.path || thumb.referenceImage?.url),
    };
```

ui/src/app/shared/thumbnail-image/thumbnail-image.directive.spec.ts
```diff
@@ -0,0 +1,276 @@
...
+      readonly rootMargin = '200px';
+      readonly callback: IntersectionObserverCallback;
+      readonly disconnect = vi.fn();
+      readonly observe = vi.fn();
+
+      constructor(callback: IntersectionObserverCallback) {
+        this.callback = callback;
+        observers.push(this);
+      }
+    }
```
victor-paunescu, line 73: `FakeIntersectionObserver` previously hardcoded `readonly rootMargin = '200px'` and ignored constructor options, causing `expect(observers[0].rootMargin).toBe('200px')` to pass unconditionally regardless of the options supplied by `ThumbnailImageDirective`. Capturing `options?.rootMargin` in the mock constructor ensures the directive's configuration is genuinely tested.

```suggestion
    class FakeIntersectionObserver {
      readonly rootMargin: string;
      readonly callback: IntersectionObserverCallback;
      readonly disconnect = vi.fn();
      readonly observe = vi.fn();

      constructor(
        callback: IntersectionObserverCallback,
        options?: IntersectionObserverInit,
      ) {
        this.callback = callback;
        this.rootMargin = options?.rootMargin ?? '';
        observers.push(this);
      }
    }
```

victor-paunescu (changes requested):

ui/src/app/shared/thumbnail-image/thumbnail-image.directive.ts
christophervoelpel: Addressed in 727f81a: moved the load binding to the directive's `host` property and removed the redundant `standalone` flag.

ui/src/app/shared/thumbnail-image/thumbnail-image.directive.ts
christophervoelpel: Addressed in 727f81a: removed the method-level decorator; `onLoad()` is now bound only through the directive's `host` property.

ui/src/app/shared/thumbnail-image/thumbnail-image.directive.ts
christophervoelpel: Added the `naturalWidth > 0` guard in 727f81a, so a zero-width load cannot hide the fallback. Regression tests cover valid and zero-width loads.

ui/src/app/shared/thumbnail-image/thumbnail-image.directive.ts
christophervoelpel: Addressed in 727f81a. The new lease stays locally owned until the DOM assignment succeeds; failures release it and retain bounded retries. The previous active lease is released only after a successful replacement, so a failed assignment cannot revoke the still-displayed image. Regression tests cover repeated assignment failures, lease cleanup and recovery.

ui/src/app/services/media/media-cache-engine.ts
christophervoelpel: Addressed in 727f81a: metadata reads now use `Promise.all`, preserving request/response ordering and the existing expiry and eviction policy. A regression test verifies that the reads start concurrently.

ui/src/app/homepage/homepage.ts
christophervoelpel: Addressed in 727f81a: reference visibility now requires a non-empty `path` or `url`; empty reference objects render the placeholder. Added tests for empty objects, empty/undefined fields and valid path/URL references.

ui/src/app/shared/thumbnail-image/thumbnail-image.directive.spec.ts
christophervoelpel: Addressed in 727f81a: the fake observer now captures `options?.rootMargin`, so the `200px` assertion checks the option actually supplied by the directive.

victor-paunescu (approved):

Issue #188, opened 2026-09-12
Speed: Project analysis


christophervoelpel:
### The Executive Verdict - Projects
  The single biggest performance issue inside a project is media re-fetching and uncompressed payloads, rather than computational complexity.
  Currently, Storyboard and Composition treat media assets in silos: Storyboard downloads and caches candidate videos locally in the browser's Cache Storage, but Composition completely ignores that cache and re-
  streams the raw clips over the internet from Google Cloud Storage. Meanwhile, the backend sends multi-megabyte uncompressed JSON payloads containing base64 thumbnails on every project load.

  Addressing these specific disconnects yields dramatic speedups across all three views.
  ──────
  ### 1. Composition Page (/:id/composition)
  #### The #1 Fix: Leverage CandidateVideoCacheService in the Composition Player
  • The Problem:
  In PR #185, CandidateVideoCacheService was added to cache candidate video bytes in the browser's Cache API (scene-machine-candidate-videos-v1). Storyboard uses this to play videos instantly.
  However, in composition.ts:247-275, heldVideoSrc resolves to remote GCS signed URLs:
    // composition.ts: resolves directly to https://storage.googleapis.com/...
    return this.mediaService.getCachedUrl(ref.path);
  When a user finishes selecting candidates in Storyboard and clicks over to Composition to preview the full video, the player re-streams every single video clip over the internet from GCS, ignoring the fact that
  those video files were already downloaded and cached locally moments earlier in the Storyboard! This leads to buffering pauses, seek delays, and jerky clip transitions.
  • The Fix:
  Have Composition resolve the current and next playlist clips via CandidateVideoCacheService.acquire(...) (or fallback to GCS if uncached).
  • Impact: Instant, 0ms-buffering playback and scrub response in Composition, because the clips are played straight out of local disk/memory blob URLs.
  ──────
  ### 2. Storyboard Page (/:id/storyboard)

  #### A. Add Batch URL Pre-signing to Storyboard
  • The Problem:
  homepage.ts:100 batch pre-signs thumbnails via MediaService.signUrls(paths), and composition.ts:288 batch pre-signs playlist videos.
  Storyboard has zero batch pre-signing.
  When a storyboard with 6 scenes and 3–4 candidates per scene opens (18–24 candidates total), every single uncached candidate thumbnail directive fires its own independent HTTP GET request to /api/signUrl?path=....
[...]

PR #189, opened 2026-09-12
Improve homepage and project-page loading

christophervoelpel:

Current head: `dd79d78`. Fresh integrated validation at #192 `dbe6f74`: **653 UI tests in 42 files**, compile, lint, spec typecheck, status-viewer tests and production build passed on Node 22.22.3/npm 10.9.3. The identical final backend tree passed **678 tests** on Python 3.13; action signatures, ShellCheck, Bash syntax and diff checks passed. The sandboxed build aborted; the same production build passed outside the sandbox.

Fixed the shared thumbnail directive, not just homepage startup: an already-acquired image stays visible while identical media gains its initial cache scope. The old lease stays alive until replacement or teardown; real media/project/resolved-bucket changes still clear stale content. Rendered tests cover pending replacement, failure/retry, destruction and timer cancellation. Reference-only homepage tiles show the normal placeholder while config loads. Added missing types to the project-list helper. The fix was reproduced failing before implementation; no new cache architecture was introduced.

## Summary

Consolidated homepage and project-page performance work for #188, stacked on
#187. This broadens the existing homepage PR rather than opening a duplicate.

- Batch first-scene Firestore reads for the project list. Join by document path
  and preserve project order, user filtering, response shape and missing scenes.
- Privately cache content-hashed Angular JS/CSS; keep HTML revalidation.
- Resolve near-viewport thumbnail URLs only after a local byte-cache miss,
  sharing signing batches bounded by both 100 paths and 3,500 characters in
  the encoded relative URL. This covers Homepage, Storyboard and Composition.
- Wait for global config to settle before initializing homepage cached images,
  including a usable fallback when config fails.
- Reuse warm candidate video bytes in Composition, with ordinary streaming for
  cold clips. Use actual image thumbnails in its filmstrip when available;
  retain the video-frame fallback for older clips without thumbnails.
- Enable opt-in gzip for successful SPA and project list/detail GETs, with
  explicit MIME types, a 500-byte minimum, negotiation and range/validator
  handling. No blanket compression of other API/worker responses.
- Document cache behavior and delivery boundaries.

## Earlier feature verification

- Node 22.22.3 / npm 10.9.3: 591 UI tests in 39 files; compile, lint, spec
  typecheck, production build and status-viewer tests pass.
- Python 3.13: 669 tests plus 15 subtests; locked dependency validation passes.
- Three bounded implementation/review agents plus integration review. Feedback
  on rapid source changes, shared-file trims and encoding negotiation was fixed.
  Tests exercise real Composition, cache and HTTP boundaries, with real pending
  requests rather than manually toggled pending flags.
- Real Chrome, final built UI: after warming two clips in Storyboard, entering
  Composition added **zero video requests**, down from three on the baseline.
  Cross-clip seeking and playback also added no video/image/signing requests.
  Cold direct Composition streamed one active video and loaded two thumbnails.
  Legacy clips without thumbnails retain their video-frame previews.
- Real local Flask middleware: baseline built JS/CSS response bodies total
  1,545,354 bytes with identity versus 402,489 with negotiated compression;
  every body round-trips. This includes lazy routes, not just homepage startup.
- Existing homepage QA: 24-project fixture with delayed config, six visible
  images. Cold image requests fell 12 to 6, signed paths 23 to 6; warm reload
  needed no image or signing request. Full regression suite rerun for this head.
- Follow-up on the analysis comment in #188: the installed Gunicorn 23 parser
  rejects a realistic 64-path signing request before Flask (4,697-byte line,
  default limit 4,094). Encoded-length partitioning fixes the remaining gap.
  Real HTTP tests cover valid Unicode names, later-chunk joins/omissions and
  failed-chunk retry without losing successful cached URLs. Fresh Chrome checks
  retain zero extra requests on warm homepage reload and warm Composition entry.

Narrow-layout caveat: at 375px Composition's existing fixed columns collapse
the preview to zero width. It reproduces with identical bounds on the baseline;
layout is explicitly out of scope. At 768/1024/1440px filmstrip images load.
No claim that phone layout is fixed or accepted by this performance PR.

## Assessment of #188

Accepted changes target confirmed redundant reads, requests and transferred
bytes. Batch reads reduce RPCs, not billed document reads. Cached media still
needs disk reads and decoding: no zero-buffering guarantee.

The following proposals are intentionally not part of this PR:

- Broad Storyboard deferred rendering: only the selected scene's candidates
  render today; images and hover previews already load on demand.
- Making MediaSrcPipe pure: async completion, URL expiry and retries depend on
  reevaluation; no measured CPU bottleneck justifies a rewrite here.
- Five-minute public template/announcement caching: templates already use an
  app-wide resource; longer HTTP caching would hide immediate admin edits.
- Partial autosave: same-project tab navigation does not await a full save;
  changing PATCH deletion/concurrency semantics deserves separate design.
- Layout changes or skeleton cards.

The follow-up analysis's Firestore ordering/fake/missing-document concerns,
hashed-file/fallback guards, config-failure fallback, dependency hash-locking
and Composition lease cleanup are already covered. Keep cache-miss-first
thumbnail signing: project metadata still refreshes and changed thumbnails
use new content-addressed paths. Signing the old path would not invalidate
its cached bytes. Keep announcement `no-store`; an ETag would still require
the Firestore read and the Markdown-only ID is not a full-payload validator.
The scoped project gzip responses do not copy incoming IAP assertion headers
into their bodies; no token exposure was reproduced from that data flow.
Production compression capacity and arbitrary stored-data side channels are
not claimed to be exhaustively validated by local tests.

No schema migration or cloud deployment. Local browser fixtures isolate cache
behavior; they are not a Cloud Run latency benchmark. #188 remains open for
review of its deferred proposals.

## Merge order

Merge after #187. Once the base lands, rebase only this PR's commits onto main,
retarget, and rerun full CI. Do not merge the stack out of order.

## Earlier stack update — 13 September

Rebased only to inherit the review fixes in ancestor PRs. A range-diff confirms this PR's own feature commits are unchanged; no additional finding or feature work was added here. Fresh combined-stack validation and exact head are recorded in the publication follow-up below. Application CI still needs to run after the parent merges and this PR is retargeted to `main`.


ui/src/app/homepage/homepage.html
```diff
@@ -84,7 +84,7 @@ <h1 class="main-title">Scene Machine</h1>
               }
 
               <!-- Fallback: Placeholder -->
-              @if (thumb.showPlaceholder) {
+              @if (thumb.showPlaceholder || (thumb.showReference && !thumbnailImagesReady())) {
```
victor-paunescu: When a project card possesses a high-quality thumbnail but lacks a low-quality preview thumbnail (common for generated video scenes), `thumb.showPlaceholder` evaluates to `false` and `thumb.showReference` evaluates to `false`.

While global configuration is loading (`!thumbnailImagesReady()`), line 65 suppresses the high-resolution image, but line 87 also fails to trigger because neither `thumb.showPlaceholder` nor `thumb.showReference` is true. As a result, the card thumbnail container renders completely blank (showing neither image nor icon) until configuration resolves.

Updating the fallback check to display the placeholder whenever high-resolution images are not yet ready and no low-quality thumbnail is available ensures cards maintain a consistent visual placeholder:

```suggestion
              @if (thumb.showPlaceholder || (!thumbnailImagesReady() && !thumb.lowQualityThumbnail)) {
```

ui/src/app/composition/composition.ts
```diff
@@ -517,6 +676,16 @@ export class Composition {
...
 
+  private haveSameSource(
+    first: MediaRef | undefined,
+    second: MediaRef | undefined,
+  ): boolean {
+    return (
+      this.sourceKey(this.candidateCacheScope(), first) ===
+      this.sourceKey(this.candidateCacheScope(), second)
+    );
+  }
```
victor-paunescu, line 688: `sourceKey(scope, ref)` returns an empty string `""` when `ref` is undefined. Consequently, if both `first` and `second` are undefined, `haveSameSource(undefined, undefined)` evaluates to `true`.

Adding an explicit guard ensures that missing media references are never treated as sharing an underlying video stream:

```suggestion
  private haveSameSource(
    first: MediaRef | undefined,
    second: MediaRef | undefined,
  ): boolean {
    if (!first || !second) return false;
    return (
      this.sourceKey(this.candidateCacheScope(), first) ===
      this.sourceKey(this.candidateCacheScope(), second)
    );
  }
```

ui/src/app/services/media/media-cache-engine.ts
```diff
@@ -126,6 +126,30 @@ export class MediaCacheEngine {
     }
   }
 
+  async acquireCached(
+    scope: MediaCacheScope,
+    file: MediaRef | null | undefined,
+  ): Promise<MediaCacheLease | null> {
+    if (!file?.path) return null;
```
victor-paunescu: In `acquire()`, persistent caching is bypassed if `scope.bucket` or `scope.projectId` is missing:
```ts
if (!persist || !file?.path || !scope.bucket || !scope.projectId)
```
In `acquireCached()`, only `!file?.path` is checked. If called with an uninitialized or empty scope (e.g. `{bucket: '', projectId: ''}` before global config loads), `this.cacheRequest()` constructs a malformed key with empty path segments (`.../v1///path`) and queries `cache.match()` needlessly.

Adding explicit validation for `scope.bucket` and `scope.projectId` avoids unnecessary storage lookups on empty scopes:

```suggestion
    if (!file?.path || !scope.bucket || !scope.projectId) return null;
```

victor-paunescu (changes requested):

ui/src/app/homepage/homepage.html
christophervoelpel: Fixed in 3ac004f. High-resolution-only cards now show the placeholder while config loads; an existing low-quality thumbnail remains visible. The rendered tests cover both cases and confirm no premature cache acquisition. Full #189 suite: 627 tests passed.

ui/src/app/composition/composition.ts
christophervoelpel: Added the missing-reference guard in 3ac004f. Missing references no longer compare equal; the existing shared-source seek regression remains green. This is boundary hardening rather than a reproduced playback failure.

ui/src/app/services/media/media-cache-engine.ts
christophervoelpel: Fixed in 3ac004f. acquireCached now checks bucket and projectId before constructing a cache request or accessing storage, matching acquire. Regression coverage verifies both incomplete scopes return null without opening Cache Storage.

victor-paunescu (dismissed):

charlie6 (approved):

christophervoelpel:
## Recovery of the remaining reviewed stack

This PR now also carries the missing #191/#192 changes in bed0b9e62cb3087618ca24bf73c058836f5202c6.
#190 was already included here. The manual merges of #191 and #192 updated
their intermediate feature branches after the earlier PRs had merged, so those
changes had not yet reached this PR or main.

The recovery adds the previously reviewed page-scoped project loading, uploaded
image previews, new-project defaults and bounded workflow polling, including
their review fixes and regression tests. It introduces no new behavior beyond
the reviewed stack.

The complete tracked tree is identical to the previously reviewed final #192
head, 536ddd18bb489df764c3679339c996bbb8457fd2 (also preserved by merge f065961).
This is one additive recovery commit; existing #189 history is retained.

Fresh local validation: 696 UI tests in 42 files, compile, lint, spec typecheck,
status-viewer tests and production build passed (Node 22.22.3 / npm 10.9.8).
The backend passed 682 tests plus 15 subtests on Python 3.13.7 after installing
the locked dependencies into a fresh environment. Action signatures,
ShellCheck and Bash syntax checks passed. Independent integration verification
found no differences from the complete reviewed snapshot. No new deployment
or live provider/browser validation was performed.

Fresh GitHub CI is green on bed0b9e, including UI build/lint/tests, Python
3.11/3.12/3.13, deploy checks and CLA; only conditional zizmor jobs were skipped.

Ready for Victor's re-review of this updated head. No merge has been performed;
main remains unchanged by this recovery.

uenke (approved):

PR #190, opened 2026-09-12
Make dictation controls compact and storyboard prompts resizable

christophervoelpel:

Current head: `1eea1e5`. Fresh integrated validation at #192 `dbe6f74`: **653 UI tests in 42 files**, compile, lint, spec typecheck, status-viewer tests and production build passed on Node 22.22.3/npm 10.9.3. The identical final backend tree passed **678 tests** on Python 3.13; action signatures, ShellCheck, Bash syntax and diff checks passed. The sandboxed build aborted; the same production build passed outside the sandbox.

Removed the unused duplicate `start()` method and changed tests to click the real microphone button. Added an external-field-change panel-close assertion and proved it fails when the guard is removed. The approved appearance, immediate recording, Stop/Cancel/Undo behavior, and layout are unchanged. Candidate-only switches for the scene Prompt still preserve dictation; actual owner/text changes can cancel it.

## Summary

Keep the idle dictation microphone compact and make the Storyboard Prompt
vertically resizable without moving the trim controls.

Following user testing, dictation now starts on the first microphone click.
Its recording/transcription/result controls sit inside the field instead of
in a floating modal. This is a UI follow-up stacked on #189; it does not
change the provider, backend, media caching or deployment configuration.

## Behavior

- Click the bottom-right microphone to start recording (subject to browser
  microphone permission). Cursor/selection is captured at that click; without
  a focused field, the transcript is appended.
- A square Stop button finishes recording and transcribes it. X cancels
  unfinished dictation. Transcription status remains visible after Stop.
- The completed result stays visible with Undo and X. Closing the result
  leaves the inserted text intact; the next mic click starts a new take.
- The active status row and textarea have separate layout boxes, so scrolling
  text and the scrollbar cannot pass behind the controls. When idle, only a
  narrow right-side mic gutter remains. No dictation backdrop blocks scrolling.
- All six dictation fields use the shared control. Existing owner/revision,
  cancellation, caret and safe-Undo guards remain in place.
- The Storyboard Prompt retains its 136px minimum and native vertical resize.
  Growing it expands the scrollable page without stretching the trim controls.
- README, developer guide and UI README describe the new interaction.

## Earlier UI validation

- Pinned Node 22.22.3/npm 10.9.3. Final combined stack at #192 `036d086`:
  **647 UI tests in 42 files**, compile, lint, spec typecheck, production build
  and status-viewer tests pass. PR190 head: `fe57414`.
- Shared control: **31 tests**, including real rendered clicks, unflushed
  transcription HTTP, pending microphone permission, cancellation, selection,
  safe Undo and terminal success/error focus. Only recorder/microphone/HTTP
  boundaries are faked.
- Earlier Chrome responsive checks at 1440x900, 1024x768, 768x1024 and 375x812. Actual
  clicks cover immediate start, pending status, cancellation, retained result,
  Undo, cursor insertion, long text, page scrolling and Edit dialog behavior.
  Phone checks use the existing collapsed navigation.
- Actual pointer resize during recording grows the prompt from 136px to 256px while
  trim positions stay fixed. Text and status have separate bounds; padding-only
  overlap and a resize-blocking CSS override were caught in Chrome and fixed.
- Independent shared-control review completed. Final Chrome error-path retest
  confirms focus remains on Details; success focus remains on Undo.
- #191 and #192 are rebased to inherit this change. Their own feature commits
  are unchanged by range-diff; no backend files change in this follow-up.

Chrome checks use a synthetic recorder and local HTTP fixture, not a real
microphone or paid provider. No cloud deployment or production data was used.
The local fixture is not included in the PR. Application CI still needs to run
after the parent merges and this PR is retargeted to main.


ui/src/app/shared/dictation/dictation-control.ts
```diff
@@ -331,14 +433,21 @@ export class DictationControl implements OnChanges, OnDestroy {
         this.recovery.set(undefined);
         this.insertionSelection = undefined;
         this.service.dismiss(this.lastOwner || this.ownerKey);
+        this.panelOpen.set(false);
+        this.focusTarget = 'trigger';
```
victor-paunescu: When a transcript is inserted, `panelOpen` remains `true` to present the "Undo" option. If the user then focuses the textarea and types a character, `ngOnChanges` detects the value update (`!selfWrite && !stillInserted`) and unconditionally sets `this.focusTarget = 'trigger'`. The subsequent `afterRenderEffect` immediately runs and calls `element.focus()` on `.dictation-trigger`, stealing keyboard focus away from the active textarea on the user's very first keystroke.

Focus should only be returned to `'trigger'` if the dictation control or one of its descendants currently owns focus.

```suggestion
        this.panelOpen.set(false);
        const owned =
          document.activeElement instanceof Node &&
          this.hostElement.nativeElement.contains(document.activeElement);
        if (owned) {
          this.focusTarget = 'trigger';
        }
```

ui/src/app/shared/dictation/dictation-control.ts
```diff
@@ -257,8 +336,27 @@ export class DictationControl implements OnChanges, OnDestroy {
...
+        const element = this.hostElement.nativeElement.querySelector(
+          selector,
+        ) as HTMLButtonElement | null;
+        if (!element || (target === 'panel' && !panelOpen)) return;
+        element.focus();
+        if (target === 'trigger' || status !== 'permission') {
+          this.focusTarget = undefined;
+        }
+      },
+    });
```
victor-paunescu, line 377: There are two concerns with this effect:

1. **Angular Render Phase Contract**: Calling `element.focus()` modifies DOM focus state and dispatches blur/focus events. In Angular (`afterRender` / `afterRenderEffect`), DOM mutations and focus changes must occur in `write` or `mixedReadWrite`, whereas `read` is strictly reserved for non-mutating layout measurements (`getBoundingClientRect`, `scrollTop`).
2. **Focus Hijacking Across Permission Prompts**: Because `this.focusTarget` is not cleared while `status === 'permission'`, if the user focuses a field while waiting for browser microphone permission, resolving permission transitions status to `'recording'` and abruptly steals focus to `.dictation-panel button`.

Using `write` and checking whether the control owns focus before claiming focus prevents unexpected focus jumps.

```suggestion
    afterRenderEffect({
      write: () => {
        const target = this.focusTarget;
        const panelOpen = this.panelOpen();
        const status = this.state().status;
        if (!target) return;
        const selector =
          target === 'panel' ? '.dictation-panel button' : '.dictation-trigger';
        const element = this.hostElement.nativeElement.querySelector(
          selector,
        ) as HTMLButtonElement | null;
        if (!element || (target === 'panel' && !panelOpen)) return;

        const activeElement = document.activeElement;
        const controlOwnsFocus =
          activeElement instanceof Node &&
          this.hostElement.nativeElement.contains(activeElement);
        const focusUnassigned =
          !activeElement || activeElement === document.body;

        if (
          target === 'panel' &&
          status === 'recording' &&
          !controlOwnsFocus &&
          !focusUnassigned
        ) {
          this.focusTarget = undefined;
          return;
        }

        element.focus();
        this.focusTarget = undefined;
      },
    });
```

ui/src/app/shared/dictation/dictation-control.ts
```diff
@@ -48,61 +49,120 @@ import {
...
+          <button
+            class="dictation-trigger"
+            type="button"
+            mat-icon-button
+            [attr.aria-label]="triggerLabel()"
+            [disabled]="isDisabled()"
+            (mousedown)="$event.preventDefault()"
+            (click)="handleTrigger()"
+          >
+            <mat-icon>mic</mat-icon>
```
victor-paunescu: Unlike all other icon buttons in this template (lines 93, 101, 110, 126, 146), this `<mat-icon>` omits `aria-hidden="true"`. Without `aria-hidden="true"`, screen readers may announce the icon ligature text ("mic") alongside or in place of the button's `aria-label`, failing accessibility checks.

```suggestion
            <mat-icon aria-hidden="true">mic</mat-icon>
```

ui/src/app/storyboard/storyboard.scss
```diff
@@ -1274,11 +1284,27 @@
...
+            bottom: 4px;
+            z-index: 2;
+          }
+          &.dictation-enabled ::ng-deep app-dictation-control.dictation-open {
+            position: relative;
+            inset: auto;
+            width: 100%;
+            align-self: stretch;
             flex: 0 0 auto;
           }
```
victor-paunescu, line 1309: 1. **Nested `::ng-deep`**: Line 1263 already defines an outer `::ng-deep { mat-form-field { ... } }`. Declaring `::ng-deep app-dictation-control` inside it produces malformed nested `::ng-deep` selectors (`mat-form-field ... ::ng-deep app-dictation-control`).
2. **Inline Style Specificity**: When a user drags the vertical resize handle, the browser writes an inline style (`style="height: ...px;"`) directly onto the `<textarea>`. Because inline styles have specificity `(1, 0, 0, 0)`, `height: auto` without `!important` cannot override the inline height when dictation opens, preventing the textarea from collapsing to accommodate the dictation panel.

```suggestion
          &.dictation-enabled:has(app-dictation-control.dictation-open)
            textarea {
            height: auto !important;
            padding-block-end: 0 !important;
          }
          &.dictation-enabled app-dictation-control {
            position: absolute;
            inset-inline-end: 16px;
            bottom: 4px;
            z-index: 2;
          }
          &.dictation-enabled app-dictation-control.dictation-open {
            position: relative;
            inset: auto;
            width: 100%;
            align-self: stretch;
            flex: 0 0 auto;
          }
```

ui/src/app/storyboard/edit-candidate-dialog.ts
```diff
@@ -114,15 +114,36 @@ import {DictationConfig} from '../shared/dictation/dictation';
...
+          z-index: 2;
+        }
+
+        &.dictation-enabled ::ng-deep app-dictation-control.dictation-open {
+          position: relative;
+          inset: auto;
+          width: 100%;
+          align-self: stretch;
+          flex: 0 0 auto;
+        }
```
victor-paunescu: `app-dictation-control` is a direct child element within `EditCandidateDialog`'s own template (line 74), meaning it automatically carries the component's encapsulation attribute. Applying `::ng-deep` strips view encapsulation and causes these positioning rules to leak into other dialogs and components. Scoping directly to `app-dictation-control` (matching `setup.scss` lines 707 and 715) preserves proper component isolation.

```suggestion
        &.dictation-enabled app-dictation-control {
          position: absolute;
          inset-inline-end: 16px;
          bottom: 4px;
          z-index: 2;
        }

        &.dictation-enabled app-dictation-control.dictation-open {
          position: relative;
          inset: auto;
          width: 100%;
          align-self: stretch;
          flex: 0 0 auto;
        }
```

victor-paunescu (changes requested):

ui/src/app/shared/dictation/dictation-control.ts
christophervoelpel: Fixed in 7cf766e. An external text edit only requests control focus if the control actually owns it. The regression types through the bound textarea, updates its revision and flushes the rendered effect; focus stays in the textarea. It fails with the original control restored. Full restacked #190 suite: 637 tests passed.

ui/src/app/shared/dictation/dictation-control.ts
christophervoelpel: Fixed in 7cf766e. Focus writes now run in the write phase. The initial explicit action can move focus into the panel, but permission resolution checks current ownership and cannot reclaim focus from another field. Tests also use the real Material form-field wrapper to cover its click-driven input focus, preserving explicit Stop/Cancel/Undo focus transfers. All 34 dictation tests pass; restoring the old control fails the typing and delayed-permission regressions.

ui/src/app/shared/dictation/dictation-control.ts
christophervoelpel: Added explicit aria-hidden="true" to the microphone mat-icon in 7cf766e, consistent with the other decorative control icons.

ui/src/app/storyboard/storyboard.scss
christophervoelpel: Fixed in 7cf766e: removed the redundant nested ::ng-deep and added the open-panel height override. Chrome verification at 1024px used a real textarea drag: its inline height became 277px, then the open panel computed 136px, with the panel directly below the textarea and no overlap. The saved resize height remains available when the panel closes.

ui/src/app/storyboard/edit-candidate-dialog.ts
christophervoelpel: Removed the unnecessary ::ng-deep from both direct-child dictation selectors in 7cf766e. The edit dialog retains its own scoped positioning; Chrome verification at 768px confirmed the dialog and microphone fit and the control remains absolutely positioned.

ui/src/app/shared/dictation/dictation-control.ts
christophervoelpel: Follow-up in b391d3d: the real-input regression now explicitly checks that Undo is available after insertion and unavailable after the user types, while textarea focus is preserved. This is test-only; dictation behaviour is unchanged. Full #190: 637 UI tests, compile, lint and spec typecheck passed on Node 22.22.3/npm 10.9.3.

@victor-paunescu, this follow-up is ready for review.

ui/src/app/setup/setup.scss
```diff
@@ -691,9 +691,15 @@
...
     min-height: 0;
     overflow: auto;
+    padding-inline-end: 64px !important;
     scrollbar-gutter: stable;
   }
 
+  &.dictation-enabled:has(app-dictation-control.dictation-open) textarea {
+    height: auto;
+    padding-block-end: 0 !important;
+  }
```
victor-paunescu, line 701: Because the Setup guidance textarea also supports native vertical resizing (`resize: vertical;` on line 683), dragging the handle applies an inline `style="height: ...px;"` attribute. To ensure the textarea collapses properly when dictation opens (matching the updates in `storyboard.scss` line 1294 and `edit-candidate-dialog.ts` line 122), add `!important` to `height: auto`.

```suggestion
  &.dictation-enabled:has(app-dictation-control.dictation-open) textarea {
    height: auto !important;
    padding-block-end: 0 !important;
  }
```

victor-paunescu (approved):

PR #191, opened 2026-09-13
Load page-scoped project data and cache uploaded image previews

christophervoelpel:
## New-project defaults follow-up — 14 September

Current head: `df39cc4`.

Added the requested shared new-project defaults: Gemini Omni (`gemini-omni-1.1-flash-preview`), 720p, audio ON, four-second candidates, and four candidates. Setup and Storyboard already consume the same ConfigService defaults; no page-specific override or saved-project migration was added.

The deployment template still seeds its model/region through `VEO_MODEL` / `VEO_REGION`. The validator now accepts Veo or Omni while retaining allowlist and region checks. Existing administrator choices are respected: to adopt Omni with an older `config.txt`, set `VEO_MODEL=gemini-omni-1.1-flash-preview` and `VEO_REGION=global`. README and DEVELOPING document this. Existing saved projects keep their settings.

Regression coverage reads the actual seed and shell templates, verifies Omni/global acceptance and regional rejection, and exercises new/reset defaults plus saved-project loading through the shared service.

Validation: integrated #192 `2b40948` passed **654 UI tests in 42 files**, compile, lint, spec typecheck and production build on Node 22.22.3/npm 10.9.3. The identical backend tree passed **681 tests plus 15 subtests** on Python 3.13.7. Focused ConfigService tests: **58 passed**. Deploy ShellCheck, rendered configuration/Bash syntax, Python syntax and whitespace checks passed. Independent review found no blocker.

This is local QA, not a new deployment or fresh upstream application CI. No paid provider calls, cloud changes, reviewer requests or messages were made. Existing reviewer assignments remain unchanged. Full application CI remains required after parent merge/rebase/retarget to main. The sections dated 13 September below are historical evidence.

Reviewed R4 head: `97ab4dd`. Fresh integrated validation at #192 `dbe6f74`: **653 UI tests in 42 files**, compile, lint, spec typecheck, status-viewer tests and production build passed on Node 22.22.3/npm 10.9.3. The identical final backend tree passed **678 tests** on Python 3.13; action signatures, ShellCheck, Bash syntax and diff checks passed. The sandboxed build aborted; the same production build passed outside the sandbox.

Added missing types to the editor writer; saves and data contracts are unchanged. Rebased to inherit #189/#190 fixes and adapted the reference-only placeholder test to the actual `ProjectSummary` response. Standalone #191: **643 UI tests in 41 files**, lint and spec typecheck passed at this head. The two rebase conflicts preserved both features (typed helper/docstring and independent homepage tests).

## Summary

Follow-up to #188, stacked on #190. This groups the page-data boundary with the image-preview behavior it
needs. Merge after #190; rebase/retarget to `main` and run fresh CI when the
parent stack lands.

- Homepage gets a small `ProjectSummary`, not a partial editable project:
  card metadata, resolved thumbnail materials, and archive/cache policy.
  The Firestore root query selects only those metadata fields and retains the
  existing single batch for first-scene reads.
- Storyboard, Composition and Output request `?view=editor`: all scenes,
  candidates and settings remain available, but Setup-only `inputConfig`
  stays out of the response. Setup loads the full project before enabling its
  form, with a recoverable loading/error state.
- Projected editor PATCH uses field updates that never write `inputConfig`.
  A concurrent Setup save therefore cannot be overwritten by a stale copy of
  those inputs. Full detail GET/PATCH keeps the existing contract.
- New Setup image uploads/pastes and Storyboard reference uploads get optional
  small, content-addressed display previews. They use the existing thumbnail
  cache (1 MiB per image, 64 MiB total, seven days). Generation continues to use
  the original image reference and original pixel dimensions.

No full-project cache, precomputed Firestore summaries, schema migration,
provider implementation changes, cache-budget changes or layout redesign.

## Safety and compatibility

- Missing editor `inputConfig` means unloaded, not empty. It must not cause
  Setup to save defaults while hydration is pending.
- Existing per-project save serialization and stale-load protection remain;
  regressions cover pending editor saves during Setup hydration, A-B-A late
  responses, full-load failure/retry, and newly created projects.
- Preview failure leaves the original usable. Replacement/removal/navigation
  fence late preview work; a new reference cannot retain the old preview.
  Original dimensions still drive crop/outpaint decisions.
- Storyboard generation carries an uploaded preview only when its output image
  path exactly matches that input. Transformed crop/outpaint outputs do not
  inherit a preview of different image content.
- Legacy image refs are not rewritten. They enter the existing bounded cache
  where possible; oversized legacy originals retain the normal display
  fallback. This is not a backfill or an offline project store.
- Firestore field selection is on homepage root reads. Editor detail still
  reads its root document and omits inputs from the HTTP response; resolving
  a first-scene selected thumbnail still requires that scene's candidate array
  on the server. No claim of eliminating all database reads/charges.

## Local payload evidence

Same synthetic fixture through the real Flask endpoints and the existing
in-memory database boundary: one project, a ~400 KB Setup composition brief,
three scenes, nine candidates with long prompts. Uncompressed JSON bytes;
this is not a deployed latency benchmark.

| Response | Base #190 | This PR |
| --- | ---: | ---: |
| Homepage list | 419,921 B | 256 B |
| Editor detail | 459,191 B | 59,143 B |
| Setup detail | 459,191 B | 459,191 B |

All three scenes and nine candidates remain in editor and Setup detail.

## Earlier implementation validation

Validated on Node 22.22.3 / npm 10.9.3 and Python 3.13:

- UI: **624 tests in 40 files**, compile, spec typecheck and lint.
- Python CI test set (`actions/test actions_lib/test test`): **677 tests +
  15 subtests**.
- Production UI build, status-viewer tests, action-signature check,
  deployment-script ShellCheck/bash syntax, and diff whitespace check.
- Three Luna implementation agents, followed by cross-review and regression
  fixes. Stale-load regressions were also checked by removing their guard:
  they failed, then passed with the fix restored.

Existing non-blocking Python SDK deprecation warnings remain. This validation
did not run the separate Python 3.11/3.12 CI matrix.

Live Chrome/GCS smoke: **NOT TESTED**. The attempted documented localhost
real-cloud loop could not impersonate the test project's runtime account
(`iam.serviceAccounts.getAccessToken` denied). No IAM permission was changed,
and no fixture server was substituted for that proof. No paid provider calls,
deployment changes, or customer assets were used.

GitHub workflows currently target `main`; this stacked PR needs a fresh CI run
after its base is retargeted. Local QA is not a claim of green upstream CI.

Contributor contract notes are in `DEVELOPING.md` under “Page-scoped project
data” and “Thumbnail cache identity”.

Shared project-load errors and Retry are now visible in Storyboard, Composition and Output, not only Setup. The rendered Storyboard regression drives the real ConfigService through editor GET 500 → visible Retry → successful projected response; stale-load, Setup hydration and empty-default save safeguards remain covered.

Projected saves now use PATCH `/api/projects/:id/editor`, with forced editor scope and no fallback to full-document PATCH. An offline probe against the actual pre-feature Flask backend reproduced Setup-input loss through the old query-form PATCH; the dedicated path instead returned 405 with data unchanged. Current-client saves therefore fail visibly/unsaved against that old backend. The legacy query alias is retained for older preview clients: retire or refresh those older open tabs before downgrading, since a new endpoint cannot retroactively change their requests. No generic API-version framework was added.

This PR also inherits the owning #182 recording-headroom, #185 documentation and #187 thumbnail-recovery fixes through rebased ancestors. No new provider/cloud/IAP smoke or deployment is claimed.

## Previous publication validation

PR head: `31a5be8`. Fresh integrated stack at #192 `f454c25`: **645 UI tests in 42 files**, compile, lint and spec typecheck pass on Node22.22.3/npm10.9.3. Production build passes on the identical runtime tree before the final test-only addition. The final backend tree passes **678 tests plus15 subtests** on Python3.13; action signatures, status-viewer tests, ShellCheck, Bash syntax and whitespace checks pass. The separate Python3.11/3.12 CI matrix was not run locally.


AndreyMarkinPPC (approved):

ui/src/app/output-video/output-video.html
```diff
@@ -14,6 +14,14 @@
...
 -->
 
+@if (configService.projectLoadError()) {
+  <div class="loading-state">
+    <p>Could not load this project. Please try again.</p>
+    <button mat-button type="button" (click)="configService.reloadProjectConfig()">
+      Retry
+    </button>
+  </div>
+} @else {
```
victor-paunescu, line 29: In `output-video.html`, the error block was added, but the intermediate loading spinner present in `composition.html` and `storyboard.html` (`@else if (configService.projectConfig.isLoading() || !configService.projectConfig.value().id)`) was omitted. 

Without this intermediate state, navigating to output video while project data is loading or reloading causes the page to immediately render the main layout with an unhydrated project configuration, leading to UI layout shifts and interactive controls operating on empty state.

```suggestion
@if (configService.projectLoadError()) {
  <div class="loading-state">
    <p>Could not load this project. Please try again.</p>
    <button mat-button type="button" (click)="configService.reloadProjectConfig()">
      Retry
    </button>
  </div>
} @else if (configService.projectConfig.isLoading() || !configService.projectConfig.value().id) {
  <div class="loading-state">
    <mat-spinner diameter="40"></mat-spinner>
    <p>Loading project...</p>
  </div>
} @else {
```

ui/src/app/setup/setup.ts
```diff
@@ -239,6 +268,22 @@ export class Setup {
...
+    const targetAspectRatio =
+      aspectRatio === '16:9'
+        ? 16 / 9
+        : aspectRatio === '9:16'
+          ? 9 / 16
+          : undefined;
+    return targetAspectRatio
+      ? Math.abs(width / height / targetAspectRatio - 1)
+      : undefined;
+  }
```
victor-paunescu, line 288: If an image element or metadata returns a zero or negative dimension (`height <= 0` or `width <= 0`), `Math.abs(width / height / targetAspectRatio - 1)` evaluates to `Infinity` or `NaN`. 

Persisting `Infinity` inside `aspectRatioDeviation` causes serialization issues when the project configuration is saved to Firestore. Adding a non-positive guard prevents invalid numeric calculations.

```suggestion
  private aspectRatioDeviation(
    width: number,
    height: number,
  ): number | undefined {
    if (height <= 0 || width <= 0) {
      return undefined;
    }
    const aspectRatio = this.config.projectConfig.value().aspectRatio;
    const targetAspectRatio =
      aspectRatio === '16:9'
        ? 16 / 9
        : aspectRatio === '9:16'
          ? 9 / 16
          : undefined;
    return targetAspectRatio
      ? Math.abs(width / height / targetAspectRatio - 1)
      : undefined;
  }
```

ui/src/app/storyboard/storyboard.ts
```diff
@@ -1282,11 +1296,31 @@ export class Storyboard {
...
+        }
+      | undefined,
+  ): void {
+    const projectId = this.config.projectConfig.value().id;
+    if (!projectId || !reference) return;
+    const paths = new Set([reference.path, reference.preview?.path]);
+    for (const path of paths) {
+      if (path) void this.thumbnailCache.invalidateCandidate(projectId, path);
+    }
+  }
```
victor-paunescu, line 1330: `invalidateReferenceMedia()` only invalidates `reference.path` and `reference.preview?.path`. 

For pre-existing scenes or assets where `highQualityThumbnail` has its own path distinct from `referenceImage`, deleting or replacing the reference image will clear `scene.highQualityThumbnail` from the project model but leave its cached entry active inside `thumbnailCache`. Accepting and invalidating `highQualityThumbnail?.path` ensures complete cache eviction.

```suggestion
  private invalidateReferenceMedia(
    reference:
      | {
          path?: string;
          preview?: {path?: string};
        }
      | undefined,
    highQuality?: {path?: string} | undefined,
  ): void {
    const projectId = this.config.projectConfig.value().id;
    if (!projectId || (!reference && !highQuality)) return;
    const paths = new Set([
      reference?.path,
      reference?.preview?.path,
      highQuality?.path,
    ]);
    for (const path of paths) {
      if (path) void this.thumbnailCache.invalidateCandidate(projectId, path);
    }
  }
```
*(Note: At call sites like line 1300, pass `this.invalidateReferenceMedia(scene.referenceImage, scene.highQualityThumbnail);`)*

ui/src/app/storyboard/storyboard.ts
```diff
@@ -1384,32 +1425,45 @@ export class Storyboard {
...
+            scene.highQualityThumbnail = preview.preview;
+          }
         } catch (error) {
           console.error(error);
         }
       }
-      this.updateScenes(scene);
+      if (isCurrentUpload()) {
+        this.updateScenes(scene);
+      }
```
victor-paunescu: If a scene is removed from the storyboard while asynchronous upload processing (`uploadMedia()`) is awaiting, `const scene = this.config.projectConfig.value().storyboard.find(s => s.id === sceneId)` resolves to `undefined`. 

Because `this.updateScenes()` has the fallback `const updatedScene = scene || this.selectedScene()`, passing `scene === undefined` will cause `updateScenes()` to inadvertently apply updates to whatever scene is currently selected. Adding `if (scene && isCurrentUpload())` ensures we do not mutate unrelated scenes.

```suggestion
      if (scene && isCurrentUpload()) {
        this.updateScenes(scene);
      }
```

ui/src/app/services/remix-engine/remix-engine.ts
```diff
@@ -1702,9 +1702,21 @@ export class RemixEngineService {
           if (!productsToOutpaintedImages[productId]) {
             productsToOutpaintedImages[productId] = {};
           }
+          const imageIndex = Number(image.image_id) - 1;
+          const inputImage = products.find(
+            product => String(product.id) === productId,
+          )?.images[imageIndex];
+          const preview =
```
victor-paunescu, line 1709: `products.find(...)?.images[imageIndex]` safely navigates `products.find(...)`, but indexes directly into `.images` without verifying that `.images` is defined. 

If a product record has an undefined or empty `images` property, this will throw `TypeError: Cannot read properties of undefined (reading '...')` at runtime. Using optional chaining (`?.images?.[imageIndex]`) protects against unexpected product payload shapes.

```suggestion
          const imageIndex = Number(image.image_id) - 1;
          const inputImage = products.find(
            product => String(product.id) === productId,
          )?.images?.[imageIndex];
          const preview =
```

victor-paunescu (changes requested):

ui/src/app/output-video/output-video.html
christophervoelpel: Fixed in b966aee. Output now shows a loading state for a loading resource or missing project ID, with errors taking precedence. Rendered tests cover initial/refresh loading, returning to ready and simultaneous loading/error. Full restacked #191 suite: 682 tests passed.

ui/src/app/setup/setup.ts
christophervoelpel: Added the nonpositive-dimension guard in b966aee. Upload metadata tests cover zero/negative width and height independently, with the existing valid-dimension case retained. This is defensive at the calculation boundary; the normal image decoder already rejects zero dimensions.

ui/src/app/storyboard/storyboard.ts
christophervoelpel: Fixed in b966aee. Reference removal and replacement now invalidate the original path, preview path and independently stored high-quality thumbnail path through one deduplicated set. Tests cover distinct replacement paths and a high-quality thumbnail with no reference object.

ui/src/app/storyboard/storyboard.ts
christophervoelpel: Fixed in b966aee with the explicit scene && isCurrentUpload() guard. The regression leaves uploadMedia pending, deletes its target scene, selects another scene, then completes the upload. Without the guard it records an unwanted save; with it there is no save, preview generation or mutation of the other scene.

ui/src/app/services/remix-engine/remix-engine.ts
christophervoelpel: Added optional chaining on images in b966aee. The returned-reference mapping test verifies that a missing input image array does not prevent the generated scene from being mapped, and no input preview is reused. This hardens that lookup without changing generation input validation.

ui/src/app/storyboard/storyboard.ts
christophervoelpel: Expanded the guard in becbe04 after an independent review found the later-await window. The upload now verifies that its captured scene is still in the live storyboard immediately before updating it. The original test covered deletion during uploadMedia; three additional cases delete through the real component while thumbnail generation, Base64 conversion or preview creation is pending. Each checks the surviving scene is unchanged and no extra updateProjectConfig call occurs. Removing the new guard makes all three fail; restoring it passes all 74 Storyboard tests. Full #191: 685 UI tests, compile, lint and spec typecheck passed on Node 22.22.3/npm 10.9.3.

@victor-paunescu, the expanded fix is pushed and ready for review.

victor-paunescu (approved):

PR #192, opened 2026-09-13
Bound workflow polling for large generation batches

christophervoelpel:
## New-project defaults follow-up — 14 September

Current head: `2b40948`.

Rebased onto #191 `df39cc4` to inherit the requested new-project defaults (Omni, 720p, audio ON, four seconds, four candidates). All eight of this PR's own polling commits are unchanged by range-diff; no polling behavior was altered in this follow-up.

Validation: integrated #192 `2b40948` passed **654 UI tests in 42 files**, compile, lint, spec typecheck and production build on Node 22.22.3/npm 10.9.3. The identical backend tree passed **681 tests plus 15 subtests** on Python 3.13.7. Focused ConfigService tests: **58 passed**. Deploy ShellCheck, rendered configuration/Bash syntax, Python syntax and whitespace checks passed. Independent review found no blocker.

This is local QA, not a new deployment or fresh upstream application CI. No paid provider calls, cloud changes, reviewer requests or messages were made. Existing reviewer assignments remain unchanged. Full application CI remains required after parent merge/rebase/retarget to main. The sections dated 13 September below are historical evidence.

Reviewed R4 head: `dbe6f74`. Fresh integrated validation at #192 `dbe6f74`: **653 UI tests in 42 files**, compile, lint, spec typecheck, status-viewer tests and production build passed on Node 22.22.3/npm 10.9.3. The identical final backend tree passed **678 tests** on Python 3.13; action signatures, ShellCheck, Bash syntax and diff checks passed. The sandboxed build aborted; the same production build passed outside the sandbox.

Added a 104-pending-scene regression through the real constructor recovery effect. It holds four HTTP requests open across a full second, checks completion cleanup and the next start-rate window, then switches projects with active and queued work and verifies cancellation/no late starts. Independently weakening either the in-flight or start-rate cap makes this test fail. Polling runtime code is unchanged; this is coverage plus inherited fixes.

## Summary

Keep small generations responsive without issuing one unconstrained status
request every three seconds for every pending scene in a large batch.

- One browser-local status budget: at most four request starts per rolling
  second and four requests in flight. It covers initial checks and retries.
- New executions first check after three seconds. Persisted executions being
  resumed can check immediately through the same budget. Healthy non-terminal
  responses wait three seconds before the next check.
- Requests for an execution are sequential; a slow HTTP response is not
  canceled and replaced by a timer tick. The selected Storyboard scene gets
  priority; other eligible requests use FIFO order.
- Transient errors use bounded exponential backoff with small deterministic
  jitter, honoring a longer Retry-After. Keep the existing single IAP refresh
  episode and ten-minute overall timeout/pending-marker recovery.
- Changing projects cancels queued, active and backoff work, releases slots
  and removes canceled queue entries. Earlier eligible work reschedules the
  wake timer. Storyboard clears its foreground hint on destroy.

No generation/Cloud Tasks concurrency changes, provider ETA prediction,
SSE/WebSocket service, batch-status endpoint, cross-tab coordinator, new
dependency, first-playback/range change or layout change.

## Tradeoffs and assumptions

The budget is per tab, not a server-wide quota. Multiple tabs/users multiply
it. Small active sets retain the three-second cadence after responses; network
latency and browser timer throttling can add delay. With roughly122 pending
executions, background completion can take roughly one queue sweep to notice
(about30seconds with fast responses); the selected scene is prioritized.
This is an explicit bulk-work tradeoff, not a blanket30second small-job poll.

The fixed4/s budget is a conservative starting policy, not a measured optimal
cloud limit. The ten-minute timeout includes waiting in the queue and keeps
pending markers for reopening. This PR does not speed up generation itself.

## Earlier implementation validation

Earlier head `f454c25`, Node 22.22.3 / npm 10.9.3:

- 645 UI tests in 42 files, compile, lint and spec typecheck passed. Production
  build passed on the identical runtime tree before the final test-only commit.
- Inherited backend: 678 tests plus 15 subtests on Python 3.13; action signatures,
  status-viewer tests, ShellCheck, Bash syntax and diff whitespace passed.
- Independent challenger: no actionable runtime finding. A suspected execution
  map collision was refuted by the active-ID guard and production caller guards.
- Sustained simulator: 122 pending jobs produced 712 status requests in 180 seconds;
  after draining to 3 jobs, each returned to a 3-second cadence. The first-sweep
  test also verifies foreground priority and the rolling request-rate limit.
- Slow-request, project-change/backoff cancellation, Retry-After and existing
  IAP/terminal-output/timeout paths remain covered. Test fixtures were updated
  to await the intentional 3-second first poll and to restore fake timers.

These use fake time and the HTTP boundary, not paid generations or production
latency measurements. Python 3.11/3.12 CI matrix and deployed load testing were
not run. Local QA is not a claim of fresh stacked application CI.

## Stack

Depends on #191 (`perf/page-scoped-projects-20260913`). The review fixes for
#182/#185/#187/#191 remain in their owning PRs and are inherited here. Review
this PR against its stacked base. After its parent merges, rebase only this
slice onto `main`, retarget and run fresh upstream CI; the current repository
workflows run for PRs targeting `main`.

No deployment, provider call, microphone access, email or reviewer-group
message is included. Victor is the requested reviewer.


ui/src/app/services/remix-engine/remix-engine.ts
```diff
@@ -98,6 +104,194 @@ class PollTimeoutError extends Error {
...
+    }
+    for (const item of this.active.values()) {
+      if (item.projectId !== projectId) {
+        item.subscriber.error(new ProjectChangedError());
+        item.subscription?.unsubscribe();
+        this.finish(item);
+      }
+    }
+    this.pump();
+  }
```
victor-paunescu, line 206: Calling `item.subscriber.error(...)` synchronously triggers the Observable teardown callback (lines 168–176), which invokes `this.pump()` for unstarted items. Because subsequent items in `this.queue` have not yet been marked `cancelled` or `finished` by the outer loop, the nested `pump()` can evaluate `!item.canRun()` and emit an error on those items prematurely, leading to duplicate error emissions when the outer loop reaches them.

Marking all stale queued items cancelled/finished and compacting the queue before notifying subscribers prevents re-entrant scheduling, and snapshotting `active.values()` avoids in-place Map mutation during iteration:

```suggestion
  setCurrentProject(projectId: string): void {
    const queueToCancel = this.queue.filter(item => item.projectId !== projectId);
    for (const item of queueToCancel) {
      item.cancelled = true;
      item.finished = true;
    }
    this.compactQueue();

    const activeToCancel = [...this.active.values()].filter(
      item => item.projectId !== projectId,
    );
    for (const item of activeToCancel) {
      item.subscriber.error(new ProjectChangedError());
      item.subscription?.unsubscribe();
      this.finish(item);
    }

    for (const item of queueToCancel) {
      item.subscriber.error(new ProjectChangedError());
    }
    this.pump();
  }
```

ui/src/app/services/remix-engine/remix-engine.ts
```diff
@@ -98,6 +104,194 @@ class PollTimeoutError extends Error {
...
+        error: error => {
+          item.subscriber.error(error);
+          this.finish(item);
+        },
+        complete: () => {
+          item.subscriber.complete();
+          this.finish(item);
+        },
+      });
+    }
```
victor-paunescu, line 238: [not sure about this one – check if true]

If `item.request()` throws synchronously (for example, if URL creation fails, configuration is unexpectedly absent, or an interceptor throws before returning an Observable), the error escapes `pump()`. Because `this.active.set(item.executionId, item)` has already been called, `finish(item)` is never reached, permanently leaking 1 active slot in `this.active` and blocking future requests for that `executionId`.

Wrapping the subscription invocation in a `try...catch` block ensures that synchronous errors trigger `this.finish(item)` and release the concurrency slot:

```suggestion
      this.active.set(item.executionId, item);
      this.starts.push(Date.now());
      item.started = true;
      try {
        item.subscription = item.request().subscribe({
          next: response => item.subscriber.next(response),
          error: error => {
            item.subscriber.error(error);
            this.finish(item);
          },
          complete: () => {
            item.subscriber.complete();
            this.finish(item);
          },
        });
      } catch (error) {
        item.subscriber.error(error);
        this.finish(item);
      }
```

ui/src/app/services/remix-engine/remix-engine-polling.spec.ts
```diff
@@ -0,0 +1,319 @@
...
+    const promises = Array.from({length: 5}, (_, i) =>
+      service.pollWorkflow(`execution-${i}`, 'project-1'),
+    );
+    await vi.advanceTimersByTimeAsync(0);
+    expect(startedAt).toHaveLength(0);
+    await vi.advanceTimersByTimeAsync(2999);
+    expect(startedAt).toHaveLength(0);
+    await vi.advanceTimersByTimeAsync(1);
+    expect(startedAt).toHaveLength(4);
+    expect(new Set(startedAt.slice(0, 3)).size).toBe(1);
```
victor-paunescu: Line 88 asserts that 4 requests started at `t = 3000ms`. Slicing `startedAt.slice(0, 3)` verifies only 3 of the 4 items in the Set. Slicing `(0, 4)` verifies that all 4 starts occurred at the exact same timestamp:

```suggestion
    expect(startedAt).toHaveLength(4);
    expect(new Set(startedAt.slice(0, 4)).size).toBe(1);
```

ui/src/app/services/remix-engine/remix-engine-polling.spec.ts
```diff
@@ -0,0 +1,319 @@
...
+    vi.useFakeTimers();
+    const status = new Subject<{sink: {output: object}}>();
+    httpClientMock.get.mockReturnValue(status.asObservable());
+    const poll = service.pollWorkflow('stale-execution', 'project-1');
+
+    await vi.advanceTimersByTimeAsync(3000);
+    projectConfig.set({id: 'project-2', storyboard: []});
+    status.next({sink: {output: {}}});
+
+    await expect(poll).rejects.toThrow('Project changed');
```
victor-paunescu, line 304: In `RemixEngineService`, project changes are captured by an Angular `effect()` that invokes `setCurrentProject` and `projectChanges.next(projectId)`. In test environments, signal effects do not fire without an explicit tick.

Without `TestBed.tick()`, the effect never executed, and the test only passed because line 287 emitted a mock response that hit `tap()`'s synchronous project comparison. Adding `TestBed.tick()` (matching lines 272 and 309) properly flushes the effect and verifies that the scheduler proactively cancels active work:

```suggestion
    await vi.advanceTimersByTimeAsync(3000);
    projectConfig.set({id: 'project-2', storyboard: []});
    TestBed.tick();

    await expect(poll).rejects.toThrow('Project changed');
```

victor-paunescu (changes requested):

ui/src/app/services/remix-engine/remix-engine.ts
christophervoelpel: Addressed in cefcadb. Stale queued entries are all marked and removed before notification can re-enter pump; active cancellation iterates a snapshot. One qualification from the probe: RxJS already prevents duplicate terminal errors reaching consumers, and canRun prevents old-project starts. The reproduced issue was redundant calls on stopped subscribers. Batching removes that redundancy while preserving cancellation behavior.

ui/src/app/services/remix-engine/remix-engine.ts
christophervoelpel: Added a small defensive guard in cefcadb using defer(item.request), routing a synchronous factory exception through the existing error/finish path. The public pollWorkflow regression throws once at the HTTP mock, then retries the same execution successfully; it fails without the wrapper. The current real HttpClient path reports errors through its Observable, so this is factory-contract hardening rather than a confirmed production incident.

ui/src/app/services/remix-engine/remix-engine-polling.spec.ts
christophervoelpel: Corrected to slice(0, 4) in cefcadb, so the assertion checks all four initial starts.

ui/src/app/services/remix-engine/remix-engine-polling.spec.ts
christophervoelpel: Fixed in cefcadb: the test now calls TestBed.tick() after changing the project and expects rejection without emitting a status response. This exercises proactive effect/scheduler cancellation. Polling tests: 10 passed. Final integrated stack: 693 UI tests, 682 backend tests, compile, spec typecheck, lint and production build passed locally.

ui/src/app/services/remix-engine/remix-engine.ts
christophervoelpel: Follow-up in 536ddd1 documents why cancellation marks/removes stale queued work before notifying subscribers and snapshots active work. This is explicitly defensive ordering: canRun and RxJS already prevent stale starts and duplicate terminal notifications reaching consumers. No polling behaviour or timing changed in this follow-up, and no private scheduler API was exposed just to test internal call order.

Restacked onto the #190/#191 follow-ups without changing this PR's earlier commits (range-diff unchanged). Full integrated tip: 696 UI tests, compile, lint and spec typecheck passed on Node 22.22.3/npm 10.9.3.

@victor-paunescu, this is ready for review.

victor-paunescu (approved):

PR #196, opened 2026-09-21
Use an explicit MIME allow-list and support AVIF

christophervoelpel:
### Summary

This PR addresses two MIME type handling issues in `actions_lib/gemini.py` (`get_mime_type`):

1. **User-visible bug (SM-3):** The UI uploaders (`ui/src/app/setup/setup.html` and `ui/src/app/storyboard/storyboard.html`) advertise `.avif` in their `accept` attributes (`.avif,.jpg,.jpeg,.png,.webp`), but the backend rejected `.avif` files with `ValueError: File extension 'avif' (...) is not allowed.`.
2. **Latent fragility (SM-4):** `get_mime_type` previously relied on `mimetypes.types_map['.' + file_extension]`. `mimetypes.types_map` is a mutable, OS-dependent, lazily-populated dictionary that does not contain `.mkv` by default (causing `KeyError: '.mkv'`), is non-deterministic (can be mutated or warmed unpredictably), and fails in minimal container images like `python:3.13-slim` without `/etc/mime.types`.

Both issues are eliminated together by replacing the `allowed_types` list and `mimetypes.types_map` lookup with an explicit, immutable module-level mapping (`ALLOWED_MIME_TYPES`) specifying canonical MIME types for all supported extensions, including `avif` (`image/avif`) and `mkv` (`video/x-matroska`).

### Changes
- Replaced `allowed_types` and `mimetypes.types_map` in `actions_lib/gemini.py` with an immutable `ALLOWED_MIME_TYPES` mapping proxy.
- Removed unused `mimetypes` import from `actions_lib/gemini.py`.
- Expanded unit tests in `actions_lib/test/test_gemini.py` to verify:
  - All allowed extensions map to their expected MIME type
  - `.avif` and `.mkv` specifically
  - Extension case-insensitivity (e.g. `.PNG`, `.AvIf`, `.MKV`)
  - Disallowed extensions raise `ValueError` with the expected message format


christophervoelpel:
## Review: merge after one 10-minute check

Multi-agent review (reviewer → independent critique agent re-verifying each claim against the code). Findings below are the ones that survived verification; recommendations the critique pass rejected are listed explicitly so they don't get re-litigated.

**The production diff is right.** One immutable table replacing a list plus a non-deterministic `mimetypes.types_map` lookup is exactly the smallest fix, the `mimetypes` import is gone, and no new abstraction or defensive branching came with it. Call sites are unchanged in signature and return type, so nothing downstream breaks. Tightening the disallowed-extension test to `assertRaisesRegex` on the real message is good — that's behaviour, not implementation.

One note on the PR body's reasoning: on Python 3.13 `mimetypes.types_map['.avif']` **is** populated, while `.mkv` is **not**. So the `KeyError: '.mkv'` claim is real, but the AVIF rejection came from the `allowed_types` list, not from `mimetypes`. Two correct fixes, slightly conflated justifications.

### The one open question that decides this PR

Nothing converts AVIF before it reaches Vertex. Verified end to end:

- `ui/src/app/setup/setup.html:200` and `ui/src/app/storyboard/storyboard.html:920` accept `.avif`
- `ImageImportService` passes files through the existing upload path unchanged
- `actions/generate_storyboard.py:326` sends the raw GCS URI via `Part.from_uri(mime_type=get_mime_type(...))`
- `actions_lib/image_converter.convert` is only reachable through the `convert_image` **workflow action** (`actions/convert_image.py:68`, declared in `ui/definitions/actions.json`) — it is not on the upload path

So AVIF bytes do reach Vertex. Current Vertex Gemini docs list images as PNG / JPEG / WEBP / HEIC / HEIF, and video as mp4 / mpeg / mov / avi / webm / flv / wmv / 3gpp — neither `image/avif` nor `video/x-matroska` appears. If that holds, this PR converts a clean, actionable `ValueError` into an opaque `INVALID_ARGUMENT` from Vertex at `generate_storyboard.py:326` and `actions_lib/gemini.py:128`.

This is docs-verified, not run-verified — I could not call the API.

**Suggested action:** push one real `.avif` GCS URI through `generate_storyboard` before merging. If Vertex rejects it, the lean fix is deleting `.avif` from the two `accept` attributes (2 lines), **not** adding a conversion hop to the upload path.

`.mkv` is unreachable from the UI either way (no `accept` attribute allows it), but the mapping does fix a real `KeyError`, so it's worth keeping as-is.

### Suggested deletions (~9 test lines, no coverage lost)

1. **`test_get_mime_type_all_allowed_extensions`** — drop the line

   ```python
   self.assertEqual(dict(gemini.ALLOWED_MIME_TYPES), expected_mappings)
   ```

   It restates the production table and asserts equality with it, so it fails on any legitimate future addition while proving nothing about behaviour. The `for ext, expected_mime in expected_mappings.items()` loop immediately below already covers every extension through the public function. It also forces `ALLOWED_MIME_TYPES` to be part of the module's public API purely for the test.

2. **`test_get_mime_type_avif_and_mkv_specifically`** — delete the method (~8 lines). Both `avif` and `mkv` are keys in `expected_mappings` and are already asserted by that loop.

- **Don't replace `MappingProxyType` with a plain dict.** It's one import and zero runtime cost, and immutability is precisely the property that kills the `mimetypes.types_map`-mutability class of bug this PR exists to fix.
- **Don't add an image-conversion step to the upload path** to make AVIF work. If Vertex rejects it, remove it from `accept` instead.
- **Don't touch** the second, semantically different `get_mime_type` in `util/gcs_wrapper.py:264` (it reads `blob.content_type`, used by `actions/edit_video.py:67` and `actions/generate_video.py:60`). Pre-existing duplicate-name smell, not worsened here.

christophervoelpel:
## Consolidated Review

**Verdict:** Ready to merge. This note reconciles the 19:50 comment on this PR against head `baf13ef`, and every item below was re-verified by tracing callers end to end and running the tests and probes listed under Evidence.

### Do before merge
Nothing beyond the normal final-head CI and reviewer approval. The change in `actions_lib/gemini.py` (an immutable `ALLOWED_MIME_TYPES` mapping replacing the old list plus `mimetypes.types_map` lookup) keeps `get_mime_type`'s signature, return type, and error format unchanged; both callers (`actions_lib/gemini.py:130`, `actions/generate_storyboard.py:326`) are untouched.

### Optional, does not block
- Drop the `assertEqual(dict(gemini.ALLOWED_MIME_TYPES), expected_mappings)` line in `test_get_mime_type_all_allowed_extensions` and delete `test_get_mime_type_avif_and_mkv_specifically` in `actions_lib/test/test_gemini.py:29-59`. Both are redundant with the extension loop at lines 46-50, which already covers avif and mkv through the public function.
- Soften the PR body's "user-visible bug (SM-3)" framing. The shipped UI (`setup.ts` `processFiles`, `storyboard.ts` `processFile`/`onDrop`/paste, all image-import paths) re-encodes any non-JPEG/PNG file to JPEG via canvas before upload, so clicking, dropping, or pasting an AVIF through the app today never hits this path. The fix mainly protects non-UI and workflow-input callers, plus `.mkv` robustness.
- The mapping fix still earns its place: workflow JSON and `cli.py` callers can hand a `.avif` or `.mkv` GCS URI straight to `actions_lib.gemini.prompt`, and the live probe shows Vertex accepts both types, so those callers now get a working call instead of a `ValueError` or `KeyError`.

- The 19:50 comment's claim that `ImageImportService` "passes files through the existing upload path unchanged," implying raw AVIF bytes reach Vertex from the UI. Every entry point (file picker, drag and drop, paste, URL and base64 import, storyboard reference upload) resolves back into `processFile`/`processFiles`, which converts to JPEG via `client-media.ts`'s `convertImage`, a genuine canvas re-encode that fails closed if the browser cannot decode the source. No UI path reaches `uploadMedia` with an unconverted AVIF file.
- The 19:50 comment's "open question that decides this PR": whether Vertex rejects `image/avif` or `video/x-matroska`. The live probe confirms both are accepted: AVIF succeeded via `from_uri` and inline bytes, MKV succeeded via inline bytes (the one `from_uri` MKV attempt hit a 429, not a 400, and an identical rerun succeeded). The suggested "delete `.avif` from the `accept` attributes if Vertex rejects it" fallback is moot.

### Evidence
- `TZ=UTC python -m pytest -q actions_lib/test/test_gemini.py` from `wt/pr196`: 19 passed, run independently twice.
- `git diff --stat origin/main..HEAD`: only `actions_lib/gemini.py` and `actions_lib/test/test_gemini.py` changed.
- Hand-reconstructed base `get_mime_type` executed directly: `.avif` raises `ValueError` before any `mimetypes` lookup, `.mkv` raises `KeyError` at the lookup, confirming both pre-fix failure modes.
- Head `get_mime_type` executed directly for `.avif`, `.mkv`, and a bogus extension: returns `image/avif`, `video/x-matroska`, and raises the expected `ValueError`.
- Line-by-line trace of every UI upload entry point in `setup.ts`, `storyboard.ts`, `client-media.ts`, and `image-import.ts`, plus `grep accept` on both HTML templates: JPEG re-encoding is universal, `accept` attributes are advisory only.
- Live Vertex probe (gemini-3.8-flash, global): AVIF via `from_uri` and inline bytes both succeeded; MKV via inline bytes succeeded; the one MKV `from_uri` failure was a 429, not a MIME rejection.
- Not run: live deployment of this branch, an end-to-end browser upload of an actual AVIF file, and any paid calls beyond the eight probe calls (two were reruns after a 429 rate limit).

### Way forward
- Confirm manually that the PR body's SM-3 description is updated (or accepted as-is) before merge, since the underlying code fix does not depend on that wording.
- Apply the two test deletions in the same PR or as a fast follow, whichever is less friction for the author.

uenke (dismissed):

christophervoelpel:

I addressed the review items recorded in the [Consolidated Review](https://github.com/google-marketing-solutions/scene-machine/pull/196#issuecomment-5767224530) on head `e9fd413e9bf8`.

- Removed the redundant MIME assertions and the dedicated AVIF/MKV test; the public extension loop still covers those mappings and the complete supported set.
- Corrected the PR description to describe the shipped UI's JPEG re-encoding and the remaining workflow/CLI scope accurately.

Verification: `TZ=UTC python -m pytest -q actions_lib/test/test_gemini.py` passed with 18 focused tests after the cleanup. The explicit mapping, case handling, and invalid-extension behavior remain covered.

The AVIF/MKV live probe and UI call-site trace from the review remain supporting evidence. No deployment or paid provider call was performed for this follow-up.

actions_lib/test/test_gemini.py
```diff
@@ -26,17 +26,41 @@
...
+            "mov": "video/quicktime",
+            "avi": "video/x-msvideo",
+            "webm": "video/webm",
+            "mkv": "video/x-matroska",
+        }
+        for ext, expected_mime in expected_mappings.items():
+            self.assertEqual(
+                gemini.get_mime_type(f"gs://bucket/file.{ext}"),
+                expected_mime,
+            )
```
victor-paunescu: After removing `self.assertEqual(dict(gemini.ALLOWED_MIME_TYPES), expected_mappings)` in `e9fd413`, this loop only verifies that `expected_mappings` is a subset of `gemini.ALLOWED_MIME_TYPES`. If a new extension is added to `gemini.ALLOWED_MIME_TYPES` in the future without updating `expected_mappings`, `test_get_mime_type_all_allowed_extensions` will still pass without testing the new extension.

Asserting `self.assertCountEqual(expected_mappings, gemini.ALLOWED_MIME_TYPES)` enforces 1:1 key coverage against `ALLOWED_MIME_TYPES` without duplicating the MIME value assertions, and wrapping the check in `self.subTest(ext=ext)` isolates any per-extension failure.

```suggestion
        self.assertCountEqual(expected_mappings, gemini.ALLOWED_MIME_TYPES)
        for ext, expected_mime in expected_mappings.items():
            with self.subTest(ext=ext):
                self.assertEqual(
                    gemini.get_mime_type(f"gs://bucket/file.{ext}"),
                    expected_mime,
                )
```

victor-paunescu (changes requested):

actions_lib/test/test_gemini.py
christophervoelpel: Addressed in `b59dc83`: added `self.assertCountEqual(expected_mappings, gemini.ALLOWED_MIME_TYPES)` and wrapped the per-extension assertions in `with self.subTest(ext=ext):` (`18 passed, 12 subtests passed`). Also synced into stacked child #201 (`a864cd8`).

victor-paunescu (approved):

PR #197, opened 2026-09-21
Preserve deployed env values on redeploy and accept quoted config

christophervoelpel:
### Summary

This PR addresses **SM-8** (MEDIUM) and **SM-9** (LOW) in `deploy.sh`:

> [!IMPORTANT]
> **The live deployment currently has only revision `00001`, so this bug is armed but has not fired yet. This PR should merge before the next redeploy.**

1. **SM-8: Preserve deployed `DICTATION_ENABLED` value across redeployments**
   - Cloud Run `--set-env-vars` replaces the container env block wholesale.
   - When `DICTATION_ENABLED` is omitted in an operator's local `config.txt` during redeployment, `deploy.sh` previously defaulted unconditionally to `1`, silently re-enabling microphone dictation even if an operator had explicitly disabled it (`0`) in a prior deployment.
   - Fix: Query the live deployed `app` Cloud Run service (`gcloud run services describe app ... --format='value(spec.template.spec.containers[0].env)'`) when `DICTATION_ENABLED` is unset in `config.txt`, extracting any deployed `DICTATION_ENABLED` value before falling back to default `1`.
   - Precedence: explicit value in `config.txt` > currently-deployed live value > default `1`. Guarded against initial deploy failures (when the service does not yet exist).
   - Keeps the guarantees documented in `README.md` and `DEVELOPING.md` true.

2. **SM-9: Accept double-quoted values in `config.txt` preflight check**
   - `config.template.txt` instructs users: *"Please use only double quotes if you need to wrap a variable below."*
   - The preflight validation regex previously required the character immediately following `=` to be `[A-Za-z0-9._$-]`, rejecting valid quoted lines such as `export PROJECT="my-project"`.
   - Fix: Update regex to `^(export )?${var}=("[^"]+"|[A-Za-z0-9._\$-]+)`, accepting double-quoted values while continuing to reject genuinely empty values like `PROJECT=` and `PROJECT=""`. Also clarify the error message to specify that double quotes should be used when wrapping values.

### Verification
- `shellcheck --severity=error deploy.sh deploy/libs.sh deploy/grant-access.sh` passed with 0 errors.
- `bash -n` syntax check passed for all deploy scripts.
- `git diff --check origin/main...HEAD` passed with 0 whitespace issues.
- `pytest actions/test actions_lib/test test` and `python -m test.test_actions_sig` passed (706 passed).


christophervoelpel:
## PR #197 — P2 — first-deploy lookup precedes API enablement

Reviewed head: `8f7399edd693f15a96910e0d3dbcba45628d28b7`.

Please defer the omitted-`DICTATION_ENABLED` live lookup at `deploy.sh:306–317` until after required API enablement, before deploying the app.

The new `gcloud run services list` call runs before this script enables `run.googleapis.com`. On a fresh project with that API disabled and the optional setting omitted, a non-interactive lookup fails and exits before reaching the advertised first-deploy default of `1`. The copied template explicitly sets `1`, so that ordinary path is unaffected; the bug concerns the documented omission path.

This conclusion comes from script ordering plus the existing list-failure test, not a live fresh-project deployment. The exact gcloud extraction formats were separately checked offline for `0`, `1`, and missing values.

Acceptance: first deploy with the optional flag omitted reaches default `1`; redeploy preserves live `0`; explicit local `0`/`1` wins; permission/network failures after provisioning still fail closed. Keep this scoped to ordering and its regression test.

christophervoelpel:
## Review: needs rework — the probe runs too early in the script

Multi-agent review (reviewer → independent critique agent re-verifying each claim against the code). Findings below survived verification; rejected recommendations are listed explicitly.

**What's good first:** the tests are not source-text greps — they extract the real block from `deploy.sh` and run it under `bash` with a stubbed `gcloud`, so precedence and every failure mode are genuinely exercised (22 pass against the branch). Quoting and `set -euo pipefail` interaction are correct throughout: `if ! VAR=$(...)` suspends errexit properly, all expansions are quoted, and the 0/1 validation runs *after* the live value is adopted, so a comma- or newline-bearing live value can't be injected into the `--set-env-vars` list at line 863.

> [!CAUTION]
> ### Blocker: breaks the first deploy on a fresh project
>
> Real line numbers on the merge result:
>
> | What | Line |
> | --- | --- |
> | New `DICTATION_ENABLED` probe (`gcloud run services list`) | ~306 |
> | Deploy-target confirmation — `read -r -p "Proceed and deploy to '$PROJECT'?"` | 390 |
> | `gcloud services enable` (incl. `run.googleapis.com`) | 448 |
> | First actual use of `DICTATION_ENABLED` | 863 |
>
> So the probe queries Cloud Run ~140 lines before the API is enabled, and ~85 lines before the operator has confirmed which project they're deploying to. `deploy.sh` explicitly supports the brand-new-project case ("on a brand-new project it is otherwise off", ~line 413).
>
> The new code is fail-closed — the test suite pins this exact path (`{"list_fail": True}` → rc 1) — so on a fresh project the deploy now aborts. I could not verify offline whether `gcloud run services list` hard-errors on a disabled API or interactively offers to enable it, but **both outcomes are wrong here**: either a first deploy that used to work now fails, or an API gets enabled on a project the user hasn't confirmed yet.

### Recommended fix

1. **Move the whole `DICTATION_ENABLED` block** to just after `gcloud services enable` (~line 455) and before the first use at 863. This is the one change that matters.

2. **Delete the env sanity probe** (commit `8f7399e`, ~16 lines at 319-334) along with it, plus its two test cases and the `full_env` / `full_env_fail` branches of the mock builder (~35 test lines):

   ```bash
   # Sanity check: the deployed app service always has env vars (e.g. ROLE, AUTH_MODE).
   # If the env list is empty, format extraction is broken; fail closed.
   ```

   It's a second full `describe` of the same resource, solely to detect that a gcloud projection *might* be wrong — defensive code for a state you can verify once by hand, at the cost of doubling preflight latency. Once the ordering in (1) is fixed, its failure mode largely can't occur. Instead, verify `env.filter(name=DICTATION_ENABLED).extract(value).flatten()` once against a live service and note the result in the PR. That manual check is the thing genuinely worth doing before merge.

3. **Reword the error text** at 313-314 / 338-339. "Set `DICTATION_ENABLED` explicitly in config.txt or check gcloud credentials" names the wrong cause for the commonest failures — API disabled, or a mistyped project. Say those instead.

### Additional finding: the widened config regex now accepts whitespace

The new pattern `("[^"]+"|[A-Za-z0-9._$-]+)` accepts *anything* inside double quotes, so:

```bash
REGION="us central1"
```

now passes config validation and fails later inside a `gcloud` flag, where the error is much harder to trace. This was caught before. `[^"[:space:]]+` inside the quoted alternative keeps the fix (quoted values) without the hole.

Related, and deliberate as far as I can tell: `PROJECT='single-quoted'` is valid bash in a sourced `config.txt` but is still rejected. The new error text tells users to use double quotes, so that's documented rather than broken — noting it only so nobody "fixes" it twice.

### Stale doc

`DEVELOPING.md:284` still reads "defaults to `1` when omitted". After this PR, omitted means "inherit the live value"; it only defaults to `1` when the service is absent. The adjacent "Existing explicit `0` values are preserved" (README.md:393, DEVELOPING.md:286) is what this PR makes true, so the scope is right — just that one sentence is now stale.

- **Don't parametrise the grep-pattern test** at `test/test_deploy_safety.py:637-646`. Scraping the pattern out of `deploy.sh` and un-escaping it is ugly, but it tests *the pattern that actually ships*; a parametrised copy duplicates deploy.sh's quoting rules and the two will drift.
- **Don't add more failure-mode test cases.** The 22 bash-level tests already cover more than this feature warrants.

christophervoelpel:
## Consolidated Review

**Verdict:** Fix one blocker, then merge. This reconciles the 19:29 and 19:50 comments on this PR against head `8f7399e`; every item below was re-verified by tracing `deploy.sh` callers and running the tests and probes under Evidence.

### Do before merge

1. Move the `DICTATION_ENABLED` preservation block (`deploy.sh:306-350`) to run after `gcloud services enable` (`deploy.sh:490`) and before its first use in `--set-env-vars` (`deploy.sh:905`). The probe fires at line 306, well before the confirmation prompt (432) and before `run.googleapis.com` is guaranteed enabled (490); nothing between there and line 905 reads its output, so relocating it changes nothing before confirmation. Acceptance: `test/test_deploy_safety.py -k dictation_env_preservation` (10 tests) still passes moved, and a first deploy with the flag omitted on a project without Cloud Run enabled no longer aborts before confirmation.

### Optional, does not block

- Tighten the quoted-value regex at `deploy.sh:267` from `[^"]+` to `[^"[:space:]]+`, so `REGION="us central1"` is caught at config-validation time instead of failing later inside a `gcloud` flag. Fails safely either way; `config.txt` is trusted, sourced input, not a security boundary.
- Update the stale sentence at `DEVELOPING.md:284` ("defaults to `1` when omitted"): omission now inherits the live value, defaulting to `1` only when the service doesn't exist yet. `DEVELOPING.md:286` and `README.md:393` are already correct.
- Reword the probe failure text at `deploy.sh:314, 325, 339` to name API enablement or an unconfirmed project, more probable than bad credentials once item 1 lands.

- The 19:50 comment's call to delete the `LIVE_ENV_VARS` sanity check (`deploy.sh:319-334`): rejected. `ROLE` and `AUTH_MODE` are always present on a real deployed service, so an empty result there means gcloud's field extraction broke, distinct from `LIVE_DICTATION` coming back empty, which is expected on a service deployed before this feature existed. Deleting the check would collapse a fail-closed signal into the same default-1 fallback as a benign case. Keep it; merging the two `describe` calls is a later optimization.
- The 19:50 comment's `PROJECT='single-quoted'` note: documented, intentional, no fix needed.
- Line numbers differ between the 19:50 comment and this note because the 19:50 comment counted on the pre-PR file; at head the probe is at 306, the confirmation prompt at 432, `services enable` at 490 and the first use at 905. Same defect either way.
- The 19:50 comment's "don't parametrize the grep-pattern test" and "don't add more failure-mode cases" notes: agreed.

### Evidence

- The three new `gcloud` projections were run unchanged against a live Cloud Run app service deployed from this repository (existing project, API enabled): the service list returned `app`, the env-name extraction returned `ROLE,AUTH_MODE,WORKER_URL,IAP_AUDIENCE,FIRESTORE_DB_UI,DICTATION_ENABLED,DICTATION_MODE`, and the `DICTATION_ENABLED` value extraction returned `1`, all with exit code 0. This is the manual format check the 19:50 comment asked for; the extraction works.
- `bash -n deploy.sh` and `shellcheck --severity=error deploy.sh`: both clean.
- `TZ=UTC python -m pytest -q test/test_deploy_safety.py`: 56 passed.
- `TZ=UTC python -m pytest -q test/test_deploy_safety.py -k dictation_env_preservation -v`: 10 passed, `list_fail` already fails closed before confirmation or API enablement.
- Regex probe: `REGION="us central1"` matches the current pattern, rejected by `[^"[:space:]]+`; `PROJECT='single-quoted'` matches neither.
- `git diff --check origin/main..HEAD`: clean.
- Not run: live deployment against a fresh project with Cloud Run disabled, browser or UI checks, paid model calls.

### Way forward

- Land the reorder plus the optional regex and doc tweaks in one follow-up commit, then re-run the affected test selections.
- Confirm manually on a fresh-project deploy (Cloud Run disabled, flag omitted) that the relocated probe runs after API enablement and lands on the default of `1`.

christophervoelpel:

I addressed the required ordering fix and the accepted optional cleanup from the [Consolidated Review](https://github.com/google-marketing-solutions/scene-machine/pull/197#issuecomment-5767224961) on head `d53ee53dd747`.

- Moved the `DICTATION_ENABLED` lookup until after confirmation and `gcloud services enable`, before the first environment-variable use.
- Kept early validation for explicit values and the fail-closed live-environment checks.
- Tightened quoted config values to reject whitespace-containing values and corrected the development-guide wording for inheritance.

Verification: `TZ=UTC python -m pytest -q test/test_deploy_safety.py` passed with 59 tests, including first-deploy ordering and preservation behavior. `bash -n deploy.sh`, ShellCheck at error severity, and `git diff --check` passed. The first-deploy sequencing is covered with mocked gcloud calls; no fresh-project deployment was performed.

deploy.sh
```diff
@@ -264,8 +264,8 @@ REQUIRED_VARS=(
 )
 MISSING=0
 for var in "${REQUIRED_VARS[@]}"; do
-  if ! grep -qE "^(export )?${var}=[A-Za-z0-9._\$-]+" ./config.txt; then
-    echo "ERROR: $var is missing, empty, or has invalid characters in config.txt" >&2
+  if ! grep -qE "^(export )?${var}=(\"[^\"[:space:]]+\"|[A-Za-z0-9._\$-]+)" ./config.txt; then
```
victor-paunescu: There are two subtle issues in the updated `grep -qE` pattern:

1. **Missing right-hand anchor and `{}` in the unquoted branch**: Because `[A-Za-z0-9._\$-]+` is not anchored to the end of the value (and does not include `{` or `}`), `grep -qE` succeeds as soon as the first character after `=` matches. For example, `export GCS_BUCKET=${PROJECT}-scene-machine` only matches the leading `$` character, and an unquoted value with spaces like `export REGION=us central1` or trailing tokens like `export PROJECT="my-project" extra` passes `grep -qE` and causes `source ./config.txt` to truncate the variable or execute the trailing token.
2. **Unrestricted punctuation inside `"[^"[:space:]]+"`**: Wrapping the template placeholder in double quotes (`export PROJECT="<YOUR_PROJECT_ID>"`, following the note in `config.template.txt`) now passes both `grep -qE` and the `[ -z "${!var}" ]` check after `source ./config.txt`.

Restricting both the quoted and unquoted branches to `[A-Za-z0-9._\${}-]+` and anchoring to the end of the line (allowing optional trailing whitespace and inline `#` comments as used in `config.template.txt`) resolves both issues while remaining compatible with the extractor in `test/test_deploy_safety.py`:

```suggestion
  if ! grep -qE "^(export )?${var}=((\"[A-Za-z0-9._${}-]+\"|[A-Za-z0-9._${}-]+)([[:space:]]*(#.*)?)?$)" ./config.txt; then
```

deploy.sh
```diff
@@ -451,6 +453,55 @@ else
...
+  if ! APP_EXISTS=$(gcloud run services list \
+    --region="$REGION" --project="$PROJECT" \
+    --filter="metadata.name=app" \
+    --format="value(metadata.name)"); then
+    echo "ERROR: Failed to query Cloud Run services in ${PROJECT}/${REGION}." >&2
+    echo "       Set DICTATION_ENABLED explicitly in config.txt or check gcloud credentials." >&2
+    echo "Validation failed. Please fix config.txt and try again." >&2
+    exit 1
+  fi
+  if [ -n "$APP_EXISTS" ]; then
```
victor-paunescu: In `gcloud run services list --filter="metadata.name=app"`, the `=` filter operator performs word-boundary token matching rather than full-string equality (so `metadata.name=app` matches other services in the same project/region containing `app` as a hyphen-delimited token, such as `my-app` or `scene-machine-app`). If such a service exists on a first deploy before `app` is created, `APP_EXISTS` will be non-empty and `gcloud run services describe app` on line 471 will fail with `NOT_FOUND`.

Checking `APP_EXISTS` with `grep -Fxq "app"` (matching the exact-line idiom used for `ENABLED_APIS` on line 436) ensures we only describe the service when `app` itself exists:

```suggestion
  if printf '%s\n' "$APP_EXISTS" | grep -Fxq "app"; then
```

test/test_deploy_safety.py
```diff
@@ -466,3 +466,278 @@ def test_announcement_seed_is_not_enabled_on_worker():
...
+        ),
+        # List-failure aborts rather than defaulting
+        (
+            None,
+            {"list_fail": True},
+            1,
+            None,
+            "Failed to query Cloud Run services",
+        ),
+    ],
```
victor-paunescu, line 587: The post-probe validation block (`deploy.sh:499-503`) is specifically kept to reject malformed `DICTATION_ENABLED` values inherited from the live `app` service (e.g., if edited manually in the Cloud Console to `"true"` or `"invalid"`), but `test_dictation_env_preservation_precedence_and_failure_modes` currently only tests live `"0"`, `"1"`, and `""`. Adding a parametrized test case for an invalid live value directly covers this fail-closed path:

```suggestion
        # List-failure aborts rather than defaulting
        (
            None,
            {"list_fail": True},
            1,
            None,
            "Failed to query Cloud Run services",
        ),
        # Invalid live DICTATION_ENABLED value on existing service aborts
        (
            None,
            {
                "list": "app",
                "full_env": "ROLE,DICTATION_ENABLED",
                "describe": "invalid",
            },
            1,
            None,
            "DICTATION_ENABLED must be 0 or 1",
        ),
    ],
```

test/test_deploy_safety.py
```diff
@@ -466,3 +466,278 @@ def test_announcement_seed_is_not_enabled_on_worker():
...
+  fi
+}}
+phase() {{ :; }}
+PROJECT=test-project
+REGION=us-central1
+REQUIRED_APIS="run.googleapis.com"
+ENABLED_APIS=
+{enable_block}
+printf 'value=%s\\n' "$DICTATION_ENABLED"
+'''
```
victor-paunescu, line 707: [Not sure about this – double check please]

Two issues in this test harness prevent it from catching regressions in `enable_block`:

1. **Missing `set -euo pipefail`**: `deploy.sh` runs under `set -euo pipefail`, and `gcloud services enable $TO_ENABLE --project=$PROJECT > /dev/null` (`deploy.sh:450`) relies on `set -e` rather than an `if !` guard to abort on failure. Without `set -euo pipefail` in `script`, if `gcloud()` hits the `else` branch and returns `1`, `bash -c` continues executing and exits with `0`.
2. **Overwritten `REQUIRED_APIS="run.googleapis.com"`**: Setting `REQUIRED_APIS="run.googleapis.com"` on line 681 has no effect because `enable_block` starts at `# --- Enable services` (`deploy.sh:410`), which immediately overwrites `REQUIRED_APIS` at `deploy.sh:418`. Checking `[[ " $* " == *" run.googleapis.com "* ]]` in the `gcloud services enable` mock branch ensures the test fails if `run.googleapis.com` is omitted from `TO_ENABLE`:

```suggestion
  script = f'''\
set -euo pipefail
gcloud() {{
  if [ "${{1:-}}" = services ] && [ "${{2:-}}" = list ]; then
    printf '%s\\n' "$ENABLED_APIS"
  elif [ "${{1:-}}" = services ] && [ "${{2:-}}" = enable ] && [[ " $* " == *" run.googleapis.com "* ]]; then
    printf 'services-enable\\n' >> "$TRACE_FILE"
  elif [ "${{1:-}}" = run ] && [ "${{2:-}}" = services ] && [ "${{3:-}}" = list ]; then
    printf 'run-list\\n' >> "$TRACE_FILE"
  else
    printf 'unexpected gcloud call: %s\\n' "$*" >&2
    return 1
  fi
}}
phase() {{ :; }}
PROJECT=test-project
REGION=us-central1
ENABLED_APIS=
{enable_block}
printf 'value=%s\\n' "$DICTATION_ENABLED"
'''
```

victor-paunescu (changes requested):

deploy.sh
christophervoelpel: Addressed in `5407608`: anchored both the quoted and unquoted branches to `[A-Za-z0-9._${}-]+` with optional trailing whitespace and `#` comment support, plus added test cases in `test/test_deploy_safety.py` covering unquoted whitespace, trailing tokens, quoted `<YOUR_PROJECT_ID>`, and inline comments.

deploy.sh
christophervoelpel: Addressed in `5407608`: updated the check to `if printf '%s\n' "$APP_EXISTS" | grep -Fxq "app"; then` and added a regression test case for prefix/substring service names (`my-app`, `scene-machine-app`).

test/test_deploy_safety.py
christophervoelpel: Addressed in `5407608`: added the parametrized test case for `"describe": "invalid"` verifying it aborts with `rc 1` and `DICTATION_ENABLED must be 0 or 1`.

test/test_deploy_safety.py
christophervoelpel: Confirmed and addressed in `5407608`: added `set -euo pipefail`, removed the dead `REQUIRED_APIS="run.googleapis.com"` assignment, and asserted `[[ " $* " == *" run.googleapis.com "* ]]` in the `gcloud services enable` mock branch (`65 passed`).

deploy.sh
```diff
@@ -264,8 +264,8 @@ REQUIRED_VARS=(
 )
 MISSING=0
 for var in "${REQUIRED_VARS[@]}"; do
-  if ! grep -qE "^(export )?${var}=[A-Za-z0-9._\$-]+" ./config.txt; then
-    echo "ERROR: $var is missing, empty, or has invalid characters in config.txt" >&2
+  if ! grep -qE "^(export )?${var}=((\"[A-Za-z0-9._{}\$-]+\"|[A-Za-z0-9._{}\$-]+)([[:space:]]*(#.*)?)?$)" ./config.txt; then
```
victor-paunescu: Because `[A-Za-z0-9._${}-]+` places `${}` inside a double-quoted bash string (`"..."`), `bash` evaluates `${}` at runtime as an empty parameter expansion before invoking `grep`, aborting with:

```
deploy.sh: line 267: ^(export )?${var}=(("[A-Za-z0-9._${}-]+"|[A-Za-z0-9._${}-]+)([[:space:]]*(#.*)?)?$): bad substitution
```

Placing `-` at the start of the bracket expression and `$` immediately before `]` (`[-A-Za-z0-9._{}$]+`) ensures `$` is followed by `]` (which `bash` treats as a literal `$` inside double quotes without requiring backslash escaping and without triggering `${}` or `$-` expansion):

```suggestion
  if ! grep -qE "^(export )?${var}=((\"[-A-Za-z0-9._{}$]+\"|[-A-Za-z0-9._{}$]+)([[:space:]]*(#.*)?)?$)" ./config.txt; then
```

test/test_deploy_safety.py
```diff
@@ -466,3 +466,314 @@ def test_announcement_seed_is_not_enabled_on_worker():
...
+{loop_block}
+exit "$MISSING"
+"""
+  proc = subprocess.run(
+      ["bash", "-c", script],
+      cwd=tmp_path,
+      capture_output=True,
+      text=True,
+      check=False,
+  )
```
victor-paunescu, line 769: Because this test extracts the regex into Python and invokes `["grep", "-qE", full_regex]` directly instead of running the double-quoted `grep -qE "..."` command through `bash -c`, it bypasses `bash` double-quote expansion and passes even when `deploy.sh` contains invalid bash parameter expansions like `${}` (`bad substitution`) or unescaped `$-`.

Executing the extracted `grep -qE "..."` command via `bash -c` under `set -euo pipefail` and asserting `proc.stderr == ""` ensures the test exercises the exact `bash` string expansion used by `deploy.sh` at runtime:

```suggestion
def test_config_validation_regex_accepts_quoted_and_rejects_empty(
    line, expected_match
):
  """Preflight regex accepts double-quoted config values but rejects empty."""
  text = _deploy_sh()
  match = re.search(
      r'(grep -qE "\^$export $\?$\{var\}=$.*?$") \./config\.txt',
      text,
  )
  assert match, "Config validation grep command not found in deploy.sh"
  grep_cmd = match.group(1)
  var_match = re.match(r"^(?:export\s+)?([A-Za-z_]+)=", line)
  var = var_match.group(1) if var_match else "PROJECT"
  script = f"set -euo pipefail\nvar={var}\n{grep_cmd}\n"
  proc = subprocess.run(
      ["bash", "-c", script],
      input=line,
      capture_output=True,
      text=True,
      check=False,
  )
  assert proc.stderr == "", f"Bash error evaluating grep command: {proc.stderr}"
  assert (proc.returncode == 0) == expected_match
```

victor-paunescu (changes requested):

deploy.sh
christophervoelpel: Addressed in `fca09b3` / `f883601`: updated `deploy.sh:267` to use `[-A-Za-z0-9._{}$]+` (with `-` immediately after `[` and `$` immediately before `]`) so `bash` treats `$` as a literal `$` inside double quotes without triggering `${}` (`bad substitution`) or `$-` shell-flag expansion.

test/test_deploy_safety.py
christophervoelpel: Addressed in `f883601`: updated `test_config_validation_regex_accepts_quoted_and_rejects_empty` to extract the `grep -qE "..."` command from `deploy.sh`, execute it through `bash -c` under `set -euo pipefail`, and assert `proc.stderr == ""` (`65 passed`). Also verified via fault-injection that `${}` fails all 18 cases immediately under `bash -c`.

victor-paunescu (approved):

PR #205, opened 2026-09-21
Clear video source on resolve failure and keep cached blobs

christophervoelpel:
## Summary

This PR addresses two LOW severity media cache robustness issues in the UI:

### Fix 1 — SM-12: Stale frame + stuck lease on resolve failure (Low Severity)
In `ui/src/app/composition/composition.ts`, `resolveCurrentSource` previously caught resolution errors from `mediaService.resolve(ref)` without updating the resolved source state. As a result, the `<video>` element remained bound to the previous clip's `blob:` URL, causing the user to see the previous scene's clip instead of entering an unresolved state, and delaying the release of the retiring lease.
- **Fix**: Added `this.setResolvedSource(epoch, key, null);` in the `catch` block when current, which resets `heldVideoSrc` to `null`, updates `sourceReady` to `false`, and triggers the immediate release of the retiring lease.

### Fix 2 — SM-13: Downloaded Blob discarded on cache write failure (Low Severity)
In `ui/src/app/services/media/media-cache-engine.ts`, `load` previously wrapped both fetch/stream reading and `enqueueMutation` (persistence to Cache Storage via `cache.put`) in a single `try/catch`. If `cache.put` failed (e.g. storage quota exceeded), the catch block caught the error and returned `{fallbackUrl}`, discarding the already-downloaded `Blob` and forcing a redundant second network fetch.
- **Fix**: Wrapped `await this.enqueueMutation(...)` in an inner `try/catch` so that storage persistence failures are safely absorbed and the in-memory `Blob` is preserved and returned to the caller as an object-URL lease.

### Notes on Refuted Claims (Not Acted On)
- **Object URL pool**: There is no 24-slot object URL pool in this codebase (`maxObjectUrls`, `acquirePlayerUrl`, etc. do not exist). Lease lifecycle is bounded and self-healing: at most one unreleased lease, which cleans up on the next switch or destroy.
- **`currentPlaylistIndex` clamping**: Not needed. Every playlist consumer already guards bounds (`composition.ts:230` bounds check, `:237` optional chaining, `:525`/`:542` `previousItem?.video`, `:566-567` early returns). Shrunk playlists degrade cleanly without crashes.
- **`thumbnail-cache.ts` batch loop**: Left untouched. The outer loop in `flushResolutions` provides failure isolation across sub-batches.

## Testing
- Extended `composition.spec.ts` with specs verifying:
  - Resolution failure clears resolved source to `null`, unbinds previous `blob:` URL, marks source not ready, and releases retiring lease.
  - Control verifying resolution success binds the new URL and releases the retiring lease.
- Extended `candidate-video-cache.spec.ts` with specs verifying:
  - `cache.put` quota rejection still returns the downloaded `Blob` lease and does not re-fetch.
  - `enqueueMutation` rejection still returns the downloaded `Blob` lease and does not re-fetch.
  - Control verifying successful cache write stores the entry and returns the blob lease.
- Updated `thumbnail-cache.spec.ts` quota rejection test to expect the retained in-memory `Blob` lease instead of discarding it.
- Gate passed: `npm run build && npm test -- --watch=false && npm run lint` (42/42 test files passed, 700/700 unit tests passed, 0 lint warnings/errors).


christophervoelpel:

Multi-agent review (reviewer → independent critique agent re-verifying each claim against the code). The critique pass **rejected** the reviewer's one substantive code suggestion; details below.

**Both production changes are one-liners in the right place, and the risky part checks out.** `setResolvedSource(epoch, key, null)` reuses the existing epoch/key fence, so a rejection arriving after navigation is ignored, and the retiring lease is then released by `releaseRetiringLeasesAfterRender` (`renderedSrc !== lease.url`) — i.e. the blob URL is revoked only *after* it is unbound from the `<video>`. `createObjectURL` (`media-cache-engine.ts:448`) stays one-to-one with `revokeObjectURL` (`:475`) via refcount, and the new path goes through the same `createLease`, so there's no leak and no growth beyond active leases.

All three specs genuinely fail against `main` (old code leaves `'blob:cached-clip1'`, and returns the signed URL instead of `'blob:lease'` / `'blob:thumbnail'`). DEVELOPING.md's "Thumbnail cache identity" rules are untouched — no change to `bucket+project+path` keying, nothing new persisted.

### Worth fixing

1. **Widen the comment at `media-cache-engine.ts:214-221`.** The new inner catch wraps `enqueueMutation(… store(…))`, and `store()` is also where the `"…too large to cache"` / `"Unable to make room…"` throws originate — so those are now swallowed too, and the retained blob sits in the JS heap (≤64 MiB per video lease; ≤1 MiB × visible thumbnails) where it previously didn't. The behaviour is right; the comment should say what it covers:

   ```
   // Storage persistence (quota, eviction, lock) failures keep the in-memory blob.
   ```

   The change is visible in a test rename (`thumbnail-cache.spec.ts:352`: `'https://signed.example/a.webp'` → `'blob:thumbnail'`), so it's worth one extra clause.

2. **`composition.ts:388-390`** checks `isCurrentSource` and then `setResolvedSource` checks it again. Harmless, but it reads as redundant — `if (!this.isCurrentSource(...)) return;` then log + set is one less level of nesting.

3. **`thumbnail-cache.spec.ts:338`** — the renamed test now asserts two unrelated behaviours ("unavailable **and** retains blob when quota rejects"). Either split it, or keep the old name and add a two-line quota test.

### Suggested deletions (~65 lines)

- **`composition.spec.ts`, "binds the resolved URL … (control)"** (~46 lines) is a near-verbatim copy of `holds the previous src (never null) across a cross-clip seek` (spec:794-839) — same mocks, same `seekTo(12)`, same resolve. Delete it and move only `expect(releaseClip1).toHaveBeenCalled()` into the existing test.
- **`candidate-video-cache.spec.ts`, "…when cache write succeeds (control)"** (~19 lines) duplicates `shares one persistent download and revokes its object URL…` (spec:85-98).

**Keep both engine throw-site tests** — they're the only coverage of the new catch. (The `enqueueMutation` one overlaps the `cache.put` one; if you want to drop one, drop that, since the `cache.put` case is the real quota scenario and asserts `cache.values.size === 0`.)

One thing worth noting: the cache behaviours are *not* asserted three times across the spec files. The overlap is within each file, not between them.

### Explicitly not recommended: don't wrap the `<video>` in `@if`

`composition.html:63` binds `[src]="heldVideoSrc()"` bare, and Angular's sanitizer stringifies `null` to `''`. Per the HTML media resource-selection algorithm an empty `src` resolves to the *document* URL, so the player can fetch the page HTML and fire a spurious `error`. That mechanism is real (pre-existing on the empty-playlist path at `composition.ts:294`).

But `@if (heldVideoSrc()) { <video …> }` is the expensive fix: `composition.ts:106` holds `viewChild('mainVideo')`, so destroying and recreating the element on every transient null drops playback position and forces every `videoElement()` consumer to handle `undefined` — a bigger correctness surface than the console noise it removes.

It's also rarer than "the normal outcome of every resolve failure": `setResolvedSource(…, null)` only fires when `isCurrentSource(epoch, key)` still holds (`composition.ts:399`), i.e. only when the user is still sitting on the clip whose signed-URL resolve just failed.

If the empty-`src` fetch ever proves to matter in practice, the cheap fix is a no-op `(error)` handler, not element recreation. Worth one manual check with DevTools Network open before merge — the claim rests on the sanitizer/HTML spec, not an observed run, and the PR's own new test asserts only `not.toContain('blob:cached-clip1')`, which passes either way.

christophervoelpel:
## Consolidated Review

**Verdict:** Ready to merge. This note reconciles the 19:50 comment on this PR against head `4ad9e98`, and every item below was re-verified by tracing the callers on both sides of the two changed code paths and running the tests and probes listed under Evidence.

### Do before merge

Nothing beyond the normal final-head CI and reviewer approval. The production diff is two one-line additions: `composition.ts`'s catch block now calls `setResolvedSource(epoch, key, null)`, and `media-cache-engine.ts` wraps the existing `enqueueMutation(...store...)` call in a local try/catch. Both are gated by the pre-existing `isCurrentSource` epoch/key fence, both are covered by new specs that fail on `main` and pass at head, and lease refcounting in `createLease`/`retain`/`release` (`media-cache-engine.ts:445-480`) stays one-to-one either way, since the new null-source path never calls `createLease` or `acquire`.

### Optional, does not block

- Widen the comment at `media-cache-engine.ts:214-221` to note that the inner catch also swallows the `"...too large to cache"` and `"Unable to make room..."` throws from `store()`, not just opaque persistence errors. Documentation only; the retained blob is still bounded by the same per-lease size caps that apply on a successful write.
- `composition.ts:388-390`: the `isCurrentSource` check before calling `setResolvedSource` is redundant with the check `setResolvedSource` runs itself. Harmless, pure style.
- `thumbnail-cache.spec.ts:337-352`: the renamed test now covers two distinct failure sites (`caches.open` rejection and a separate `cache.put` rejection) under one name. Worth splitting for clarity, not a coverage gap.
- `composition.spec.ts:888-926` and `candidate-video-cache.spec.ts` (new "(control)" tests) overlap existing specs in setup and, for the candidate-video-cache one, in assertions too. Trimmable, but for the composition.ts case any trim needs to keep the `video.src` and `sourceReady` assertions the control test adds, since the older test only checks the intermediate `heldVideoSrc` state.

- The suggestion to wrap the `<video>` element in `@if (heldVideoSrc())` was already rejected in the 19:50 comment itself and holds up: `composition.ts:106`'s `viewChild('mainVideo')` would need optional-chaining everywhere it is read if the element could be destroyed and recreated, and the null state this PR adds is transient and gated by `isCurrentSource`, not the normal outcome of every resolve failure. Nothing in this pass found reason to revisit that call.
- The empty-`src` document-URL fetch mechanism the comment flags is real per the HTML media resource-selection algorithm, but it is pre-existing: `composition.ts:294` already sets `heldVideoSrc` to `null` on the empty-playlist path, untouched by this diff. This PR only adds a second, narrower trigger. Not a defect introduced here.

### Evidence

- `git diff --check origin/main..HEAD`: clean, exit 0.
- `npm test -- --watch=false --include composition.spec.ts --include candidate-video-cache.spec.ts --include thumbnail-cache.spec.ts`: 3 files, 91 tests, lint passed. Run twice, same result.
- Read `media-cache-engine.ts:191-224`, `:313-334`, `:445-480` and `composition.ts:283-296`, `:370-403` to trace the catch sites, the `store()` throw sites, and the lease refcount invariant.
- Base-fail reasoning by diff inspection: on `main`, the one added line in each file is the only path that produces the new tests' assertions, so all three new specs are unreachable on base.
- Not run: live browser check of the empty-`src` fetch behaviour, live deployment, any paid model call.

### Way forward

- If picking up the optional cleanup items, do them as a small follow-up commit rather than reopening this PR, to keep the review scope tight.
- Confirm CI is green on the current head before merging; nothing here should have changed that.

christophervoelpel:

I reviewed the [Consolidated Review](https://github.com/google-marketing-solutions/scene-machine/pull/205#issuecomment-5767228343) against head `4ad9e9830049`. No further change was warranted: the two production additions already clear stale resolved-source state and contain cache-store failures, while the existing epoch/key fence and lease refcounting remain intact.

Verification: the focused composition, candidate-video-cache, and thumbnail-cache suites passed with 91 tests and lint. The combined branch also passed spec typecheck and production build. No live-cloud browser check, deployment or paid model call was performed.

victor-paunescu (approved):

